PR #26, opened 2024-04-18
#25 set up docker

avu-ua:
Created files for dockerization:

1. ._env - for sensitive data (to be renamed to .env before use)
2. Dockerfile
3. docker-compose.yml

Files created in the Folder "ForumProject" which is expected to be the Project's root folder.


ForumProject/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
+RUN pip install --upgrade pip
+
+WORKDIR /app
+
+COPY . /app/
+RUN apt-get update && apt-get install -y curl && apt-get clean
+RUN pip install -r requirements.txt
```
mehalyna: Before installing Python dependencies, copy only the `requirements.txt` file and install the dependencies. This utilizes Docker's cache to skip reinstalling packages if `requirements.txt` hasn't changed:
```
COPY requirements.txt /app/
RUN pip install -r requirements.txt
COPY . /app/
```

ForumProject/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+version: "3.9"
+
+services:
+  postgres_lib:
+    image: postgres:latest
```
mehalyna, line 5: Instead of using `postgres:latest`, it is advisable to pin a specific version of PostgreSQL (e.g., `postgres:16`). This prevents potential compatibility issues with future versions and ensures consistent behavior across environments

ForumProject/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+      - .env
+    ports:
+      - "5432:5432"
+
+  forum:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: forum
+    depends_on:
```
mehalyna, line 19: Although `depends_on` ensures that `postgres_lib` starts before `forum`, it doesn't guarantee that PostgreSQL is ready to accept connections when the `forum` service starts. Consider using a tool like `wait-for-it` or `dockerize` to delay the forum service startup until the database is actually ready to accept connections

mehalyna (changes requested):

ForumProject/Dockerfile
avu-ua: Done, thank you

ForumProject/docker-compose.yml
avu-ua: As discussed, I keep "latest" for now

ForumProject/docker-compose.yml
avu-ua: Thanks, this has been implemented with 'dokerize' tool

PR #35, opened 2024-04-22
#29 define django models for users

Andrewakiv:


ForumProject/users/models.py
```diff
@@ -1,3 +1,55 @@
...
+            raise ValueError('Superuser must have is_superuser=True.')
+        return self.create_user(email, password, **extra_fields)
+
+
+class CustomUser(AbstractBaseUser, PermissionsMixin):
+    id = models.BigAutoField(primary_key=True)
+    first_name = models.CharField(max_length=20)
+    last_name = models.CharField(max_length=20)
+    email = models.EmailField(max_length=100, unique=True)
+    password = models.CharField(max_length=255)
```
mehalyna: In Django, you typically don't need to define a `password` field in your user model explicitly because it is provided by `AbstractBaseUser`. Django handles password hashing and storage behind the scenes

ForumProject/users/models.py
```diff
@@ -1,3 +1,55 @@
...
+    phone_number = models.CharField(max_length=20, unique=True)
+    is_active = models.BooleanField(default=False)
+    is_staff = models.BooleanField(default=False)
+    is_superuser = models.BooleanField(default=False)
+
+    USERNAME_FIELD = 'email'
+    REQUIRED_FIELDS = ['first_name', 'last_name', 'password', 'phone_number']
```
mehalyna: The `email` and `password` are prompted for by default and should not be included in `REQUIRED_FIELDS`

ForumProject/users/models.py
```diff
@@ -1,3 +1,55 @@
...
+        return self.create_user(email, password, **extra_fields)
+
+
+class CustomUser(AbstractBaseUser, PermissionsMixin):
+    id = models.BigAutoField(primary_key=True)
+    first_name = models.CharField(max_length=20)
+    last_name = models.CharField(max_length=20)
+    email = models.EmailField(max_length=100, unique=True)
+    password = models.CharField(max_length=255)
+    phone_number = models.CharField(max_length=20, unique=True)
```
mehalyna: Consider the implications if two users can share the same phone number (for example, family members or shared company lines)

mehalyna (changes requested):

ForumProject/users/models.py
Andrewakiv: Done, thanks

ForumProject/users/models.py
Andrewakiv: Done, thanks

ForumProject/users/models.py
Andrewakiv: Done, thanks

mehalyna (approved):

PR #39, opened 2024-04-22
#11a Implementing Startup Company Profile Creation

### LINKED ISSUE #11: Implementing Startup Company Profile Creation
1. **Familiarization with Project Requirements**:
   - Review the project specifications and user stories to understand the scope and requirements.

2. **Database Model Implementation**:
   - Define database models for startup profiles, including fields for company information and projects.

3. **API Endpoint Creation for Profile Management**:
   - Develop API endpoints to handle profile creation, updating, and viewing.
   - Use Django Rest Framework's serializers and view sets for API handling.

4. **Input Validation and Data Security**:
   - Implement input validation to ensure data integrity.
   - Discuss the importance of secure data handling practices.

5. **Front-end Integration (Optional)**:
   - If applicable, create a simple front-end form to submit and edit profiles.
   - This can be a basic HTML template or a simple JavaScript-based UI.

6. **Testing and Debugging**:
   - Write tests for the profile creation functionality.

7. **Documentation and Code Review**:
   - Document the functionality and any important decisions made during development.

8. **User Story Validation**:
    - Verify that the developed feature meets the acceptance criteria outlined in the user story.
 

US #10

DariyBautista:
1. Startup deletion is possible only if the Projects related to the Startup have a certain status (for example, "closed") or if the Startup has no Projects
2. An attempt to create a Startup with an empty field name, or industry, or phone, or country, or ciy, or address (or making changes to an existing Startup so that any of these fields become empty) should produce an error with the corresponding message.
3. In startup_logo you can add only a graphic file, in startup_phone – validation by regular expression. For startup_country - a set of choices, with a list of all countries of the world
4. The startup must have a unique name, that is, an attempt to create a startup with a name that already exists in one of the existing ones must fail and return an error message


ForumProject/projects/models.py
```diff
@@ -0,0 +1,20 @@
+from django.db import models
+from django.core.exceptions import ValidationError
+from startups.models import Startup
+
+
+class Project(models.Model):
+    
+    project_name = models.CharField(max_length=150, unique=True)
+    startup_id = models.ForeignKey(Startup, on_delete=models.CASCADE, related_name='projects')
```
mehalyna: Conventionally, the name of a `ForeignKey` field is the name of the related model in lowercase (unless you have a good reason to name it differently). Instead of `startup_id`, just `startup` would be more appropriate

ForumProject/projects/models.py
```diff
@@ -0,0 +1,20 @@
...
+    PROJECT_STATUS_CHOICES = [
+        ('open', 'Open'),
+        ('closed', 'Closed'),
+        ('pending', 'Pending'),
+    ]
+    project_status = models.CharField(max_length=20, choices=PROJECT_STATUS_CHOICES, default='open')
+    id = models.AutoField(primary_key=True)
```
mehalyna: There's no need to explicitly define an `id` field as `AutoField` because Django does this automatically if no primary key field is specified. It's completely fine to include it if you want to be explicit, or if you plan to change the type of field later, but if not, you can remove it to clean up your model

ForumProject/projects/serializers.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class ProjectSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Project
+        fields = [
+            'id',
+            'project_name',
+            'project_status',
+            'startup_id',
```
mehalyna: If you followed the earlier advice and changed the `ForeignKey` field name from `startup_id` to `startup`, you need to update the serializer fields to reflect that change

ForumProject/projects/serializers.py
mehalyna, line 1: If you have any custom validation logic that needs to be applied when a Project is created or updated, you should add corresponding validator methods in your serializer. For example, if you want to validate the uniqueness of the project_name against a particular startup, you'd write a custom validation method:
```
def validate_project_name(self, value):
    if Project.objects.filter(startup=self.initial_data['startup'], project_name=value).exists():
        raise serializers.ValidationError("A project with this name already exists for the given startup.")
    return value
```

ForumProject/projects/serializers.py
```diff
@@ -0,0 +1,15 @@
...
+            
+        ]
```
mehalyna: If there are any fields that should not be modified during update operations, you should specify them using `read_only_fields` in the `Meta` class
```
class Meta:
    model = Project
    fields = ['id', 'project_name', 'project_status', 'startup']
    read_only_fields = ['id']  # Add any fields that should be read-only
```

ForumProject/projects/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.urls import path, include
+from . import views
+from rest_framework import routers
+
+app_name = 'projects'
+
+router = routers.DefaultRouter()
+router.register('', views.ProjectView)
```
mehalyna: The `router.register` method requires two arguments: a prefix for the URLs, and a viewset. It appears that you are only passing `views.ProjectView`, but you're missing the URL prefix which defines the route at which the viewset will be accessible. E.g.
```
router.register('projects', views.ProjectView)
```

mehalyna: Make sure that `views.ProjectView` is the correct name of the viewset class you intend to use. Typically, viewsets are named in a way that ends with 'ViewSet' (e.g., `ProjectViewSet`) to distinguish them from regular views

ForumProject/projects/views.py
```diff
@@ -0,0 +1,10 @@
+from django.shortcuts import render
+from rest_framework import viewsets, status
+from rest_framework.response import Response
+from .models import Project
+from .serializers import ProjectSerializer
+
+# Create your views here.
+class ProjectView(viewsets.ModelViewSet):
+    queryset = Project.objects.all()
+    serializer_class = ProjectSerializer
```
mehalyna: By default, a `ModelViewSet` does not apply any permission checks beyond authentication, which means any authenticated user could potentially perform any operation. You may want to define and apply permissions to protect your API endpoints. E.g.
```
from rest_framework.permissions import IsAuthenticated, IsAdminUser

class ProjectView(viewsets.ModelViewSet):
    # ... existing code ...
    permission_classes = [IsAuthenticated, ]  # or any other permission class
```

mehalyna (changes requested):

ForumProject/projects/models.py
DariyBautista: Done, thanks

ForumProject/projects/serializers.py
DariyBautista: Done, thanks

ForumProject/projects/serializers.py
DariyBautista: Done, thanks

ForumProject/projects/views.py
DariyBautista: These conditions were not written down in the task, as they do not apply to startups

Please review the Startup app

ForumProject/projects/urls.py
DariyBautista: Done, thanks

DariyBautista: this is how the endpoint was conceived - an empty string, if you specify startups there, I will get the url [localhost](http://127.0.0.1:8000/startups/startups)

ForumProject/projects/serializers.py
DariyBautista: Done, thanks

ForumProject/projects/models.py
DariyBautista: Done, thanks

ForumProject/projects/urls.py
DariyBautista: http://127.0.0.1:8000/startups/startups/

alona-ruda (approved):

mehalyna (approved):

PR #41, opened 2024-04-23
#5-Implement-User-Sign-Up

Andrewakiv:
Completed registation. We will set up gmail smpt later due to two-step verification problems with the new gmail address.


avu-ua (approved):

ForumProject/users/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+        phone_number = attrs.get('phone_number')
+
+        if password != password2:
+            raise serializers.ValidationError(
+                {"password": "Password fields didn't match."})
+
+        try:
+            CustomUserValidator.validate_password(password)
+        except ValidationError as error:
+            raise ValueError({'password': error.detail})
```
mehalyna: It's more appropriate to raise `serializers.ValidationError` instead of `ValueError` to keep it consistent with the DRF's way of handling validation errors

mehalyna: When raising `ValidationError`, you can pass the message directly without wrapping it into a dictionary with the field as a key. DRF will handle associating the error message with the field based on where the exception is raised

ForumProject/users/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+
+            Returns:
+            - (CustomUser): The newly created user instance.
+        """
+        user = CustomUser.objects.create(
+            first_name=validated_data['first_name'],
+            last_name=validated_data['last_name'],
+            email=validated_data['email'],
+            phone_number=validated_data['phone_number'],
+        )
```
mehalyna: It's better to use a `create_user` method (which you should define in your `CustomUserManager`) instead of `CustomUser.objects.create` for user creation. This method can handle things like setting the password correctly and performing other custom user creation logic

ForumProject/users/tests.py
```diff
@@ -1,3 +1,46 @@
...
+            'id': 1,
+            'first_name': 'Martin',
+            'last_name': 'Daniels',
+            'email': 'daniels@gmail.com',
+            'phone_number': '+380974131765'
+        }
+        url = reverse('users:register')
+        response = self.client.post(url, data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response.data, expected_data)
```
mehalyna: When testing API endpoints, especially for user creation, you should avoid directly comparing the entire response data to expected data due to potential changes in the response format, such as the addition of new fields or security measures that prevent passwords from being returned. Instead, check for the existence of certain keys or values in the response, e.g.
```
self.assertIn('first_name', response.data)
self.assertIn('last_name', response.data)
```

ForumProject/users/tests.py
```diff
@@ -1,3 +1,46 @@
 from django.test import TestCase
+from django.urls import reverse
+from rest_framework import status
+from rest_framework.test import APITestCase
 
-# Create your tests here.
+
+class PostTests(APITestCase):
+
+    def test_register_user(self):
```
mehalyna: The `test_register_user` function includes a `password` and `password2` confirmation. You should also test that the user cannot be created when the passwords do not match, and this should return a `400 Bad Request`

ForumProject/users/tests.py
mehalyna, line 1: Since your `UserRegisterSerializer` includes a unique validator for the email, you should test the behavior when trying to register a user with an email that already exists in the database. This should also result in a `400 Bad Request`

ForumProject/users/tests.py
mehalyna: Be cautious about asserting the full content of `response.data` since that may contain sensitive information. It is generally safer to test only for the presence of expected fields rather than their exact content, unless you are sure that the content is safe to expose

ForumProject/users/tests.py
mehalyna, line 1: If your tests are creating actual entries in the database, consider overriding the `tearDown` method to clean up after each test, although `APITestCase` should handle this by rolling back the database after each test

ForumProject/users/tests.py
```diff
@@ -1,3 +1,46 @@
...
+            'first_name': 'Martin',
+            'last_name': 'Daniels',
+            'email': 'daniels@gmail.com',
+            'password': 'Rootroot1$',
+            'phone_number': '+380974131765'
+        }
+
+        url = reverse('users:register')
+        response = self.client.post(url, data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
mehalyna: You should also include tests for other validation errors (like invalid email format, missing fields, etc.)

ForumProject/users/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        - user (CustomUser): The user object.
+        - token (str): The verification token.
+
+        Returns:
+        - None
+
+        Sends an email to the user's email address containing a verification link.
+        """
+        verification_link = reverse('users:email-verify', kwargs={'token': str(token)})
+        absolute_url = 'http://' + site + verification_link
```
mehalyna: Using `http://` is not secure, especially for a verification link which typically contains sensitive information. You should ideally use `https://` especially if the site supports SSL, which is standard for modern web applications

ForumProject/users/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        email = EmailMessage(
+            subject='Verify your email',
+            body='Hi ' + user.first_name + ' Use the link below to verify your email \n' + absolute_url,
+            from_email=settings.EMAIL_HOST_USER,
+            to=[user.email]
+        )
+        email.send()
```
mehalyna: The method `email.send()` can throw exceptions if there is an issue with the email backend (like a network problem or configuration error). You could wrap the `send()` call in a `try-except` block and log any exceptions or even retry the sending mechanism

ForumProject/users/views.py
```diff
@@ -1,10 +1,101 @@
...
+        - A Response object with the result of the email verification.
+            - If the token is valid and the associated user is successfully activated,
+            returns a success message with HTTP status code 200.
+            - If the token is expired, returns an error message indicating the activation has expired
+            with HTTP status code 400.
+            - If the token is invalid, returns an error message indicating an invalid token
+            with HTTP status code 400.
+        """
+        try:
+            payload = jwt.decode(token, settings.SECRET_KEY, "HS256")
```
mehalyna: Advice for the future. You hard-coded the algorithm ('HS256'). It's better practice to fetch this dynamically from your settings to ensure consistency across your application and to facilitate changes.

mehalyna (changes requested):

ForumProject/users/utils.py
Andrewakiv: Due to this problem, since we have no SSL, I can skip `http://` and use only the domain.

ForumProject/users/serializers.py
Andrewakiv: Done, thanks

Andrewakiv: Done, thanks

ForumProject/users/serializers.py
Andrewakiv: Done, thanks

ForumProject/users/tests.py
Andrewakiv: Done, thanks

ForumProject/users/tests.py
Andrewakiv: Done, thanks

ForumProject/users/tests.py
Andrewakiv: Done, thanks

ForumProject/users/tests.py
Andrewakiv: Done, thanks

ForumProject/users/tests.py
Andrewakiv: Done, thanks

ForumProject/users/tests.py
Andrewakiv: Done, thanks

ForumProject/users/utils.py
Andrewakiv: Done, thanks

ForumProject/users/views.py
Andrewakiv: Done, thanks

ForumProject/users/tests.py
mehalyna, line 1: Consider using `setUpTestData` for data that doesn't change across tests, like URL definitions. This method runs once per test case, before all tests, saving time in tests where the data setup is the same for all tests.
You might also want to ensure that your email mocking is scoped only to the tests that need it, rather than globally, to avoid side effects affecting other tests unexpectedly.
```
def setUp(self):
    super().setUp()
    self.base_data = {
        'first_name': 'Martin',
        'last_name': 'Daniels',
        'email': 'daniels@gmail.com',
        'password': 'Rootroot1$',
        'password2': 'Rootroot1$',
        'phone_number': '+380974131765'
    }
    self.url = reverse('users:register')

def test_missing_first_name(self):
    data = self.base_data.copy()
    del data['first_name']
    response = self.client.post(self.url, data, format='json')
    self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```

mehalyna, line 1: You're testing boundary cases for phone numbers. Consider also testing boundary cases for other inputs, like maximum length for names and emails, or edge cases for any other custom validation logic you might have

ForumProject/users/validators.py
```diff
@@ -0,0 +1,60 @@
...
+        if len(password) < 8:
+            raise ValidationError('Password length should be longer or equal than 8')
+        if not any(symbol.isdigit() for symbol in password):
+            raise ValidationError('Password should have at least one digit')
+        if not any(symbol.isupper() for symbol in password):
+            raise ValidationError('Password should have at least one uppercase letter')
+        if not any(symbol.islower() for symbol in password):
+            raise ValidationError('Password should have at least one lowercase letter')
+        if not any(symbol in punctuation for symbol in password):
+            raise ValidationError('Password should have at least one symbol')
```
mehalyna: The repeated scans of the password for digits, uppercase, lowercase, and symbols can be optimized by scanning the password once and counting the occurrences of each type. This reduces the computational cost, especially for long passwords:
```
digit, upper, lower, symbol = False, False, False, False
for char in password:
    if char.isdigit():
        digit = True
    if char.isupper():
        upper = True
    if char.islower():
        lower = True
    if char in punctuation:
        symbol = True
```

mehalyna: It's more clear to accumulate all errors and report them at once, rather than stopping at the first error:
```
errors = []
if len(password) < 8:
    errors.append('Password length should be at least 8 characters.')
if not digit:
    errors.append('Password should have at least one digit.')
if not upper:
    errors.append('Password should have at least one uppercase letter.')
if not lower:
    errors.append('Password should have at least one lowercase letter.')
if not symbol:
    errors.append('Password should have at least one symbol.')
if errors:
    raise ValidationError(' '.join(errors))
```

ForumProject/users/views.py
```diff
@@ -1,10 +1,101 @@
...
+
+    This view handles GET requests with an activation token in the URL.
+    If the token is valid and not expired, it activates the user associated with the token.
+
+    Attributes:
+    - permission_classes: The list of permission classes applied to this view.
+    """
+    permission_classes = [AllowAny]
+
+    def get(self, request, token):
```
mehalyna: Using a `GET` request to change server state (activating a user) is not recommended. A `POST` method would be more appropriate here.

mehalyna (changes requested):

ForumProject/users/tests.py
Andrewakiv: Done, thanks

Andrewakiv: Done, thanks

ForumProject/users/validators.py
Andrewakiv: Done, thanks

Andrewakiv: Done, thanks

ForumProject/users/views.py
Andrewakiv: Done, thanks

mehalyna (approved):

mehalyna (approved):

PR #46, opened 2024-04-23
some changes

DmAlexx:
added:
- realization sign out;
- unit test of sign out


ForumProject/users/views.py
```diff
@@ -1,11 +1,28 @@
...
+    permission_classes = [IsAuthenticated]
+
+    def post(self, request):
+        try:
+            refresh_token = request.data["refresh_token"]
+            token = RefreshToken(refresh_token)
+            token.blacklist()
+            return Response({"message": "User successfully logged out."}, status=status.HTTP_200_OK)
+        except Exception as e:
+            return Response({"error": "Invalid token."}, status=status.HTTP_400_BAD_REQUEST)
```
mehalyna: Specific exception handling could improve error granularity. For example, handling `TokenError` separately could provide a specific message for token-related issues, while other unexpected issues could be logged or handled differently

mehalyna (approved):

PR #47, opened 2024-04-23
#37 startup tests

avu-ua:


ForumProject/startups/tests.py
```diff
@@ -1,3 +1,160 @@
...
+        '''
+        # Create a CustomUser for authentication
+        self.user = CustomUser.objects.create_user(
+            email='dummy@user.com',
+            first_name='John',
+            last_name='Doe',
+            phone_number='+3801234567',
+            password='password',
+            is_active=True
+        )
```
mehalyna: You might consider using `setUpTestData` for creating data that does not change across tests, like the user object. This method runs once per class, not before each test method, potentially speeding up your tests.

ForumProject/startups/tests.py
mehalyna, line 1: There's repeated code for creating posts in multiple test methods. You could abstract this into a helper method within the test class to avoid duplication and make the tests cleaner and easier to maintain, e.g. 
```
def create_startup(self, data):
    return self.client.post(reverse('startups:startup-list'), data, format='json')

def test_ok_create_startup(self):
    response = self.create_startup(self.startup_data)
    self.assertEqual(response.status_code, status.HTTP_201_CREATED)
    self.assertEqual(Startup.objects.count(), 1)
    self.assertEqual(Startup.objects.first().startup_name, 'Django Dribblers')
```

ForumProject/startups/tests.py
```diff
@@ -1,3 +1,160 @@
...
+        self.assertEqual(Startup.objects.count(), 1)
+        
+        # Create second Startup
+        self.startup_data['startup_name'] = 'Python catchers'
+        response = self.client.post(reverse('startups:startup-list'), self.startup_data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(Startup.objects.count(), 2)
+
+        self.assertEqual(Startup.objects.filter(startup_name='Django Dribblers').exists(), True)
+        self.assertEqual(Startup.objects.filter(startup_name='Python catchers').exists(), True)
```
mehalyna: Consider using Django’s `@transaction.atomic` decorator for tests that involve multiple database operations which should be rolled back in case of an error during the test
https://docs.djangoproject.com/en/5.0/topics/db/transactions/

mehalyna (changes requested):

ForumProject/startups/tests.py
avu-ua: Done. Thank you - tests run time reduced by half!

ForumProject/startups/tests.py
avu-ua: Done

ForumProject/startups/tests.py
avu-ua: Done

ForumProject/startups/tests.py
```diff
@@ -1,3 +1,167 @@
...
+        self.assertEqual(Startup.objects.count(), 1)
+        
+        # Create second Startup
+        self.startup_data['startup_name'] = 'Python catchers'
+        response = self.create_startup(self.startup_data)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(Startup.objects.count(), 2)
+
+        self.assertEqual(Startup.objects.filter(startup_name='Django Dribblers').exists(), True)
+        self.assertEqual(Startup.objects.filter(startup_name='Python catchers').exists(), True)
```
mehalyna, line 93: The modification of `self.startup_data` directly affects subsequent tests since `setUpTestData` is only run once per class. It’s safer to copy`self.startup_data` to a local variable within each test method to avoid side effects. E.g.
```
def test_ok_create_two_startups(self):
    '''
    Tested creation of two Startups with unique names.
    '''
    # Create first Startup
    response = self.create_startup(self.startup_data.copy())
    self.assertEqual(response.status_code, status.HTTP_201_CREATED)
    self.assertEqual(Startup.objects.count(), 1)

    # Create second Startup with a different name
    modified_data = self.startup_data.copy()
    modified_data['startup_name'] = 'Python catchers'
    response = self.create_startup(modified_data)
    self.assertEqual(response.status_code, status.HTTP_201_CREATED)
    self.assertEqual(Startup.objects.count(), 2)
```

ForumProject/startups/tests.py
mehalyna, line 1: Wrapping test methods with `@transaction.atomic` ensures that any database changes are rolled back at the end of each test, keeping your test database clean. However, this is typically handled automatically by Django's `TestCase`, which wraps each test method in a transaction by default. Explicit use of `@transaction.atomic` might be redundant unless you have specific reasons for its use

mehalyna (changes requested):

ForumProject/startups/tests.py
avu-ua: I understand that even if the `self.startup_data` persists with modifications, each test works with an independent test database, and Django transactions roll back changes at the end of each test case. For this reason I see no need to use deep copies. If you insist that the issue of avoidance of side effects still needs to be attended to, I suggest then to move definition of startup_data dictionary from setUpTestData function to setUp (use of deepcopy would make the code less readable, in my view, and maybe more resourcefull). Please let me know.

ForumProject/startups/tests.py
avu-ua: Django’s `@transaction.atomic` decorator was implemented following your recommendation. I am happy to roll this change back (done).

DariyBautista (approved):

ForumProject/startups/tests.py
mehalyna: If your test needs to assert that certain operations are performed within a transaction or need to rollback part of the transaction as a test step, then using `@transaction.atomic` can be appropriate. E.g., we have `test_ok_create_two_startups()` within this TestSuite with two operations inside it. My advice was to wrap it with `test_ok_create_two_startups()` and all methods with similar behavior

ForumProject/startups/tests.py
mehalyna: The change, wich was provided by me,  ensures that the original `startup_data` is not altered, preserving the integrity of other tests that may rely on the original data
https://docs.djangoproject.com/en/5.0/topics/testing/tools/
image
We're using PostgreSQL with SQL transactions full support

mehalyna (changes requested):

ForumProject/startups/tests.py
avu-ua: I have moved the startup_data to SetUp - this way it gets regenerated at the start of each test, thereby ensuring any subsequent tests do not get affected. I think keeping startup_data in setUpTestData (purpose of which is to create the data that is supposed to remain unchanged across all the tests) only for the sake of making deep copy thereof in almost every test is less preferred route.

ForumProject/startups/tests.py
avu-ua: Ok, I wrapped the tests which provide for more than one transaction (or more than one attempt to do a transaction) with the database.

mehalyna (approved):

PR #48, opened 2024-04-24
#9 implementing password recovery

Andrewakiv:


ForumProject/users/serializers.py
mehalyna, line 1: **DRY** The password validation logic is repeated in both the registration and password reset serializers. Consider abstracting this into a utility method or a base serializer to reduce duplication and centralize the password validation logic
```
class BasePasswordSerializer(serializers.ModelSerializer):
    def validate_passwords(self, password, password2):
        if password != password2:
            raise serializers.ValidationError({"password2": "Password fields didn't match."})
        try:
            CustomUserValidator.validate_password(password)
        except ValidationError as error:
            raise serializers.ValidationError({"password": error.detail})

class UserRegisterSerializer(BasePasswordSerializer):
    # fields and other methods...
    def validate(self, attrs):
        self.validate_passwords(attrs.get('password'), attrs.get('password2'))
        return attrs

class PasswordResetSerializer(BasePasswordSerializer):
    # fields and other methods...
    def validate(self, attrs):
        self.validate_passwords(attrs.get('password'), attrs.get('password2'))
        return attrs
```

ForumProject/users/tests.py
mehalyna, line 1: **DRY** There's some repetition in setting up the data dictionary for each test. Consider using a setup method to initialize a base dictionary of valid data, and then modify only the necessary parts in each specific test

ForumProject/users/utils.py
mehalyna, line 1: **DRY** There's repetition in how emails are structured and sent in both methods. Abstract common functionality into a helper method to construct and send emails

ForumProject/users/validators.py
```diff
@@ -0,0 +1,60 @@
...
+        if len(password) < 8:
+            raise ValidationError('Password length should be longer or equal than 8')
+        if not any(symbol.isdigit() for symbol in password):
+            raise ValidationError('Password should have at least one digit')
+        if not any(symbol.isupper() for symbol in password):
+            raise ValidationError('Password should have at least one uppercase letter')
+        if not any(symbol.islower() for symbol in password):
+            raise ValidationError('Password should have at least one lowercase letter')
+        if not any(symbol in punctuation for symbol in password):
+            raise ValidationError('Password should have at least one symbol')
```
mehalyna: You might consider optimizing the password validation by iterating over the password string only once and checking for all required conditions during that single iteration

ForumProject/users/views.py
```diff
@@ -1,10 +1,143 @@
...
+
+    This view handles GET requests with an activation token in the URL.
+    If the token is valid and not expired, it activates the user associated with the token.
+
+    Attributes:
+    - permission_classes: The list of permission classes applied to this view.
+    """
+    permission_classes = [AllowAny]
+
+    def get(self, request, token):
```
mehalyna: Using a `GET` request to change server state (activating a user) is generally not recommended. A `POST` method would be more appropriate as it explicitly indicates that an operation changes the state.

ForumProject/users/views.py
```diff
@@ -1,10 +1,143 @@
...
+            payload = jwt.decode(token, settings.SECRET_KEY, settings.SIMPLE_JWT['ALGORITHM'])
+            user = CustomUser.objects.get(id=payload['user_id'])
+            if not user.is_active:
+                user.is_active = True
+                user.save()
+            return Response({'email': 'Successfully activated'}, status=status.HTTP_200_OK)
+        except jwt.ExpiredSignatureError as identifier:
+            return Response({'error': 'Activation Expired'}, status=status.HTTP_400_BAD_REQUEST)
+        except jwt.exceptions.DecodeError as identifier:
+            return Response({'error': 'Invalid token'}, status=status.HTTP_400_BAD_REQUEST)
```
mehalyna: Consider returning HTTP 404 Not Found instead of HTTP 400 Bad Request when the user does not exist, as it's a more accurate representation of the situation where a user ID does not correspond to any record

ForumProject/users/views.py
```diff
@@ -1,10 +1,143 @@
...
+
+        serializer = RecoveryEmailSerializer(data=request.data)
+        if serializer.is_valid():
+            try:
+                user = CustomUser.objects.get(email=email)
+                token = RefreshToken.for_user(user).access_token
+                Util.send_recovery_email(get_current_site(request).domain, user, token)
+                return Response({'success': 'Email was sent successfully'}, status=status.HTTP_200_OK)
+            except CustomUser.DoesNotExist:
+                return Response({'error': 'User does not exist'}, status=status.HTTP_400_BAD_REQUEST)
```
mehalyna: It might be safer to always return HTTP 200 OK regardless of whether the user exists or not, to prevent enumeration attacks where an attacker could use this feature to discover valid emails

ForumProject/users/views.py
```diff
@@ -1,10 +1,143 @@
...
+
+    def post(self, request, token):
+        serializer = PasswordResetSerializer(data=request.data)
+        if serializer.is_valid():
+            try:
+                payload = jwt.decode(token, settings.SECRET_KEY, settings.SIMPLE_JWT['ALGORITHM'])
+                user = CustomUser.objects.get(id=payload['user_id'])
+                user.set_password(serializer.validated_data.get('password'))
+                user.save()
+                return Response({'success': 'Password has been successfully updated'}, status=status.HTTP_200_OK)
```
mehalyna: You are decoding the token and resetting the password if valid. Consider also verifying the user's session or other context-specific details to enhance security

ForumProject/users/views.py
```diff
@@ -1,10 +1,143 @@
...
+                return Response({'error': 'User does not exist'}, status=status.HTTP_400_BAD_REQUEST)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+class PasswordResetView(APIView):
+    permission_classes = [AllowAny]
+
+    def get(self, request):
+        return Response({'success': 'Enter new data'}, status=status.HTTP_200_OK)
```
mehalyna: "Enter new data," might confuse clients. It's meant to display more descriptive message or handling

mehalyna (changes requested):

ForumProject/users/views.py
oleksandr-liashenko: Done

ForumProject/users/views.py
Andrewakiv: Done

ForumProject/users/views.py
oleksandr-liashenko: Done

ForumProject/users/serializers.py
Andrewakiv: Done

ForumProject/users/tests.py
Andrewakiv: Done

ForumProject/users/utils.py
Andrewakiv: Done

ForumProject/users/validators.py
Andrewakiv: Done

ForumProject/users/views.py
Andrewakiv: Done

mehalyna (approved):

PR #55, opened 2024-04-26
#15 a define permissions to user as startup and user as investor roles

Andrewakiv:
Permission for Project create method is not completed


ForumProject/investors/serializers.py
```diff
@@ -0,0 +1,41 @@
...
+    Attributes:
+        id (int): The ID of the startup (read-only).
+        startup_name (str): The name of the startup.
+        startup_logo (ImageField): The logo of the startup.
+        startup_industry (str): The industry of the startup.
+        startup_phone (str): The phone number of the startup.
+        startup_country (CountryField): The country where the startup is located.
+        startup_city (str): The city where the startup is located.
+        startup_address (str): The address of the startup.
+    """
```
mehalyna, line 18: The docstring incorrectly references the "Startup model" instead of the "Investor model". This could lead to confusion for someone reading the code. Make sure to update the docstring to accurately describe the `InvestorSerializer`

ForumProject/investors/views.py
mehalyna, line 1: _Advice for future_
Using `Investor.objects.all()` will retrieve all investor records from the database. This is generally fine, but if your database grows large, consider implementing query optimization such as selecting related fields if there are foreign keys or using a `prefetch_related` for reverse relationships to improve performance.
If you have a lot of Investor records, you might want to implement pagination to manage the size of the HTTP response.

ForumProject/users/permissions.py
mehalyna, line 1: 1) In `InvestorPermission` and `ProjectPermission`, you've hardcoded a return `False` for the 'list' action, which might not be intended for all users. Ensure this is the desired behavior.
2) For object-level permissions, ensure the checks for user associations with the given object (`obj`) are correct and match our application's models and relationships.
3) For actions like `update`, `partial_update`, and `destroy`, you need to ensure that only users with the right level of access (e.g., user&startup, user&investor) can perform these actions. This might require more complex checks than just existence within a related model. The retrieval actions (retrieve) seem to be properly handled across the permission classes, checking for user association with the relevant object. 
4) Utilize `get_queryset()` in your views to handle filtering based on permissions when applicable. This can sometimes be more efficient and easier to manage than complex permissions classes.
5) Always keep the principle of least privilege in mind. Users should only be given the minimum permissions they need to perform their actions.

mehalyna (approved):
Change-request only for docstring

ForumProject/investors/serializers.py
Andrewakiv: Done, stupid mistake, my bad

ForumProject/users/permissions.py
Andrewakiv: 1. Due to the absence of follow logic, I hardcoded this action, ensuring that all investors are hidden from users, startups, and other investors.
2. ok
3. We don't have role logic for CEO, managers, etc., so we can't perform more complex checks beyond verifying if the current user belongs to the obj from the user&startup or user&investor tables.
4. Thanks
5. ok

ForumProject/investors/views.py
Andrewakiv: Thanks. I will work on it.

PR #58, opened 2024-04-29
Project creation & update, Project logs (partial implementation)

avu-ua:
Done:
- several endpoints (work) for CRUD operations with Project, Project Files, Project Logs

TODO:
- write Documentation
- add Tests
- implement Logs creation for adding Project Files


ForumProject/projects/admin.py
```diff
@@ -1,3 +1,46 @@
...
+                    'budget_currency',
+                    'budget_amount'
+                    ]
+    search_fields = ['name', 'description']
+
+admin.site.register(Project, ProjectAdmin)
+
+class ProjectFilesAdmin(admin.ModelAdmin):
+    list_display = ['id', 'project', 'file_description', 'file']
+    search_fields = ['id', 'project', 'file_description', 'file']
```
mehalyna: Including `id` in `search_fields` is usually not necessary unless you specifically need to search by `id`. Users rarely search by `id` as it is not user-friendly.
Searching by the file path (`file`) may not be efficient or useful. Users generally won’t search by the file path (or name) unless you have a specific use case for it

ForumProject/projects/admin.py
mehalyna, line 1: Ensure that the fields listed in `search_fields` are indexed at the database level, especially for large datasets, to avoid performance issues. Using `search_fields` on fields that aren't indexed can be slow.
If there are a large number of entries in your database, using fields in `list_display` that require complex database queries (like calling methods or accessing related fields not using `select_related` or `prefetch_related`) can lead to performance issues.

ForumProject/projects/admin.py
```diff
@@ -1,3 +1,46 @@
...
+
+admin.site.register(ProjectFiles, ProjectFilesAdmin)
+
+class InvestorProjectAdmin(admin.ModelAdmin):
+    list_display = ['id', 'investor', 'project', 'share']
```
mehalyna, line 27: You might want to include fields from the related `Investor` and `Project` models in `list_display` by defining a method in the `admin` class that fetches the specific related field you want to display

ForumProject/projects/models.py
```diff
@@ -2,56 +2,181 @@
...
+        ProjectLog.objects.create(
+            project=self,
+            project_title=self.name,
+            change_date=current_date,
+            change_time=current_time,
+            user_id=1,  # This is a placeholder
+            action=action,
+            previous_state=previous_state[:ProjectLog._meta.get_field('previous_state').max_length],
+            modified_state=modified_state[:ProjectLog._meta.get_field('modified_state').max_length]
+            )
```
mehalyna: The method retrieves the original instance from the database for every update operation, which might be inefficient. Consider optimizing this process, possibly by caching frequently accessed data or limiting the database calls.
The method lacks error handling, especially when performing operations that might fail (like database queries)

ForumProject/projects/models.py
```diff
@@ -2,56 +2,181 @@
...
+        # Check for existing files with the same name and create a unique name
+        if os.path.exists(full_path):
+            base, extension = os.path.splitext(filename)
+            counter = 1
+            while os.path.exists(full_path):
+                new_filename = f"{base}_{counter}{extension}"
+                full_path = os.path.join(folder_path, new_filename)
+                counter += 1
+
+        return full_path
```
mehalyna: The method model should ideally be static or classmethod since it does not depend on instance-specific data

ForumProject/projects/models.py
```diff
@@ -2,56 +2,181 @@
...
+        filename = filename.replace(" ", "_")
+        
+        # Create a unique file path
+        full_path = os.path.join(folder_path, filename)
+
+        # Check for existing files with the same name and create a unique name
+        if os.path.exists(full_path):
+            base, extension = os.path.splitext(filename)
+            counter = 1
+            while os.path.exists(full_path):
```
mehalyna: Consider adding error handling in case the file operations (like `os.path.exists`) fail

ForumProject/projects/urls.py
```diff
@@ -8,5 +8,25 @@
...
+        'delete': 'destroy',
+    }), name='project_files_by_project'),
+    path('delete-file/<int:projectfiles_id>/of-project/<int:project>/',
+        views.delete_project_file, name='delete_project_file'),
+    path('shortlist-project/<int:project_id>/by-investor/<int:investor_id>/',
+        views.shortlist_project, name='shortlist_project'),
+    path('subscribe-for-project/<int:project_id>/by-investor/<int:investor_id>/for-share/<int:share>/',
+        views.subscribe_for_project, name='subscribe_for_project'),
+    path('delist-project/<int:project_id>/by-investor/<int:investor_id>/',
+        views.delist_project, name='delist_project'),
```
mehalyna: Consider renaming URL path segments to use dashes instead of underscores for better readability and consistency with common web practices (e.g., `delete-file` is good, `subscribe-for-project` is consistent, but `delist_project` should be `delist-project`)

ForumProject/projects/urls.py
```diff
@@ -8,5 +8,25 @@
...
+    path('shortlisted-projects-of-startup/<int:startup_id>/', 
+        views.shortlisted_projects_of_startup, name='shortlisted_projects_of_startup'),
+    path('shortlisted-projects-of-investor/<int:investor_id>/',
+        views.shortlisted_projects_of_investor, name='shortlisted_projects_of_investor'),
+    path('logs/<int:project>/', views.ProjectFilesViewSet.as_view({
+        'get': 'list',
+    }), name='project_logs')
 ]
```
mehalyna: Although it's more of a style choice, RESTful API design typically avoids using verbs in URLs (like delete, subscribe, delist) and instead relies on HTTP methods to indicate action. Consider if these can be refactored to use resource-based paths with appropriate HTTP methods. 
_Example_
```
urlpatterns = [
    path('', include(router.urls)),  
    path('projects/<int:project_id>/files/', views.ProjectFilesViewSet.as_view({
        'get': 'list',  # Get all files for a project
        'post': 'create',  # Add a new file to the project
    }), name='project_files'),

    path('investors/<int:investor_id>/projects/', views.InvestorProjectsViewSet.as_view({
        'get': 'list',  # Get all projects associated with an investor
    }), name='investor_projects'),

    path('investors/<int:investor_id>/projects/<int:project_id>/', views.InvestorProjectDetailViewSet.as_view({
        'put': 'update',  # Update details of a project for an investor
        'delete': 'destroy',  # Remove a project from an investor's list
    }), name='investor_project_detail'),

    # For subscribing or changing subscription details
    path('investors/<int:investor_id>/projects/<int:project_id>/subscribe/', views.SubscribeProjectViewSet.as_view({
        'post': 'create',  # Subscribe to a project
        'patch': 'partial_update',  # Change subscription details
        'delete': 'destroy',  # End subscription
    }), name='subscribe_project'),

    # View for accessing and managing project logs
    path('projects/<int:project_id>/logs/', views.ProjectLogsViewSet.as_view({
        'get': 'list',  # List logs for a project
    }), name='project_logs'),
]

```

ForumProject/projects/views.py
mehalyna, line 1: **DRY** There's repeated logic across APIView functions for checking object existence and handling errors. Consider using decorators or helper functions to streamline these checks.For common patterns like checking for object existence or logging actions, utility functions can reduce code duplication and simplify maintenance.
Consider splitting these views into separate files based on their purpose (e.g., projects.py, files.py, investors.py) to keep your codebase manageable and organized.

mehalyna (changes requested):

ForumProject/projects/admin.py
avu-ua: `id` and `file` removed from `search_fields`

ForumProject/projects/admin.py
avu-ua: Done, attribute `db_index=True` added to the Model where necessary

ForumProject/projects/admin.py
avu-ua: I consider this is not necessary, for admin purposes it is enough to see `__str__` representation of `Investor` and `Project`

ForumProject/projects/models.py
avu-ua: I gave it consideration and came to a conclusion that optimizing at this stage is hardly achievable. We can not know what Projects may be subject to frequent update operations, and if a Project gets updated, for proper logging purposes its original instance has to be retrieved, to create a log with "before-after" record. 
Errors handling is added. 
Please note the logic has been moved from `save()` method in models,py to views.py.

ForumProject/projects/models.py
avu-ua: I spent some time in attempts to make the method a class or static one, until I realized it does actually depend on instance-specific data, for example here `project_slug = slugify(self.project.name)[:20]`. The method has been left as is.

ForumProject/projects/models.py
avu-ua: Done

ForumProject/projects/urls.py
avu-ua: Renamings have been made

ForumProject/projects/urls.py
avu-ua: Renamings have been made

ForumProject/projects/views.py
avu-ua: This has been implemented, at least to a large extent.

mehalyna (approved):

PR #70, opened 2024-05-02
#15 implementing dual role functionality

Andrewakiv:


ForumProject/investors/serializers.py
```diff
@@ -35,26 +35,26 @@ class Meta:
...
+    #
+    #     value = value.strip()
+    #     value = value[0].upper() + value[1:]
+    #     if not value:
+    #         raise serializers.ValidationError("Investor name cannot be empty.")
+    #     if Investor.objects.filter(investor_name=value).exists():
+    #
+    #         raise serializers.ValidationError("Investor name must be unique.")
+    #     return value
+    #
```
mehalyna: It is a bad practice to push commented lines of code into the repository. It is strictly prohibited on projects

ForumProject/investors/views.py
mehalyna, line 1: For actions like `retrieve`, `update`, and `destroy`, you are checking if the `request.user.user_info.company_id` matches the `pk` provided in the URL. This is a good security practice but could lead to maintenance challenges or bugs if `user_info` or `company_id` attributes are not reliably set on every `User` object. Consider abstracting these checks into a permission class or method to reduce redundancy and centralize business logic.

The permission check `request.user.user_info.company_id != kwargs['pk']` is repeated in multiple methods. Consider creating a method or a permission class to handle this check to reduce duplication and simplify future modifications.

E.g.
```
from rest_framework.permissions import BasePermission

class IsCompanyMember(BasePermission):
    """
    Allows access only to members of the company.
    """
    def has_permission(self, request, view):
        # Adjust the logic based on your specific field and error handling needs
        try:
            return request.user.user_info.company_id == view.kwargs['pk']
        except AttributeError:
            return False

# In your viewset
class InvestorViewSet(viewsets.ModelViewSet):
    ...
    def get_permissions(self):
        # Map 'IsCompanyMember' to appropriate actions
        if self.action in ['retrieve', 'update', 'destroy']:
            permission_classes = [IsCompanyMember]
        elif self.action == 'list':
            permission_classes = [IsAdminUser]
        elif self.action == 'create':
            permission_classes = [IsInvestorRole]
        else:
            permission_classes = [IsAuthenticated]
        return [permission() for permission in permission_classes]
```

https://www.django-rest-framework.org/api-guide/permissions/

ForumProject/investors/views.py
```diff
@@ -49,3 +72,9 @@ def create(self, request, *args, **kwargs):
             return Response(serializer.data, status=status.HTTP_201_CREATED)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def destroy(self, request, *args, **kwargs):
+        if request.user.user_info.company_id != kwargs['pk']:
+            return Response({'error': 'You have no permission to delete investor'}, status=status.HTTP_400_BAD_REQUEST)
```
mehalyna: Returning a `400 Bad Request` for permission errors is not typical; `403 Forbidden` is more appropriate because it explicitly indicates that the operation is not allowed due to permissions

ForumProject/projects/views.py
mehalyna, line 1: The checks for company ownership and the user's role are repeated across methods. This could be abstracted into a permission class to reduce redundancy and simplify maintenance

ForumProject/projects/views.py
```diff
@@ -17,3 +22,49 @@ class ProjectViewSet(viewsets.ModelViewSet):
...
+        project = get_object_or_404(self.queryset, pk=kwargs['pk'])
+        if request.user.user_info.company_id != project.startup.id and request.user.user_info.role != 'investor':
+            return Response({'error': 'You have no permission to view this project'}, status=status.HTTP_400_BAD_REQUEST)
+        serializer = self.serializer_class(project)
+        return Response(serializer.data)
+
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        if request.user.user_info.company_id != instance.startup.id and request.user.user_info.role != 'investor':
+            return Response({'error': 'You have no permission to view this project'}, status=status.HTTP_400_BAD_REQUEST)
```
mehalyna: You have a copy-paste error in the error message ("You have no permission to view this project" should be "You have no permission to delete this project")

ForumProject/startups/views.py
```diff
@@ -51,7 +61,31 @@ def create(self, request, *args, **kwargs):
...
+            return Response({'error': 'You have no permission to update startup'}, status=status.HTTP_400_BAD_REQUEST)
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        self.perform_update(serializer)
+        return Response(serializer.data)
+
+    def partial_update(self, request, *args, **kwargs):
+        if request.user.user_info.company_id != kwargs['pk']:
+            return Response({'error': 'You have no permission to update startup'}, status=status.HTTP_400_BAD_REQUEST)
```
mehalyna: Consider using more appropriate HTTP status codes for permission errors. Using `status.HTTP_403_FORBIDDEN` instead of `status.HTTP_400_BAD_REQUEST` is more semantically correct for permission-related actions

ForumProject/startups/views.py
mehalyna, line 1: Given the repeated checks for company membership across methods, abstracting this into a permission class would help maintain DRY principles and ease future modifications

ForumProject/users/views.py
```diff
@@ -60,6 +66,76 @@ def post(self, request, *args, **kwargs):
...
+            user = request.user
+            if UserRoleCompany.objects.filter(user=user).exists():
+                user_role = UserRoleCompany.objects.get(user=user)
+                user_role.role = serializer.validated_data['role']
+                user_role.company_id = 0
+                user_role.save()
+            else:
+                UserRoleCompany.objects.create(user=user, role=serializer.validated_data['role'])
+            return Response({'success': 'Role has been successfully updated.'}, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
mehalyna: Instead of using `UserRoleCompany.objects.filter(user=user).exists()` and then `UserRoleCompany.objects.get(user=user)`, which makes two database queries, you could use `get_or_create()` to streamline this process

ForumProject/users/views.py
```diff
@@ -60,6 +66,76 @@ def post(self, request, *args, **kwargs):
         return response
 
 
+class RoleSelectionView(APIView):
+    permission_classes = [IsAuthenticated]
+
+    def post(self, request):
```
mehalyna, line 137: The POST method here is fine, but consider if PUT might be more appropriate if the operation is idempotent (i.e., the server state is the same no matter how many times the request is repeated with the same parameters)

ForumProject/users/views.py
```diff
@@ -60,6 +66,76 @@ def post(self, request, *args, **kwargs):
...
+                    return Response({'success': 'The investor company was successfully selected'}, status=status.HTTP_200_OK)
+
+            elif user_role_company.role == 'startup':
+                user_startups = UserStartup.objects.filter(customuser=self.request.user).values('startup')
+                if {'startup': company} in user_startups:
+                    user_role_company.company_id = company
+                    user_role_company.save()
+                    return Response({'success': 'The startup company was successfully selected'}, status=status.HTTP_200_OK)
+            return Response({'error': 'You must select only your company'}, status=status.HTTP_400_BAD_REQUEST)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
mehalyna: 1. Similar to 'RoleSelectionView', use more efficient queries. For example, instead of fetching all user's companies and then checking if a company is in that list, consider querying directly for the specific company
2. You should handle cases where `UserRoleCompany.objects.get(user=user)` might not find an entry, which will raise a `DoesNotExist` exception. Use `get_object_or_404` or handle this case explicitly.

ForumProject/users/views.py
```diff
@@ -60,6 +66,76 @@ def post(self, request, *args, **kwargs):
...
+    permission_classes = [IsAuthenticated]
+
+    def post(self, request):
+        serializer = RoleSerializer(data=request.data)
+        if serializer.is_valid():
+            user = request.user
+            if UserRoleCompany.objects.filter(user=user).exists():
+                user_role = UserRoleCompany.objects.get(user=user)
+                user_role.role = serializer.validated_data['role']
+                user_role.company_id = 0
```
mehalyna: Resetting `company_id` to 0 might be problematic if 0 is not a valid `company_id`. Consider setting it to `None` if the field allows nulls to properly denote no company is selected

ForumProject/users/views.py
```diff
@@ -60,6 +66,76 @@ def post(self, request, *args, **kwargs):
...
+
+    def get_queryset(self):
+        user = UserRoleCompany.objects.get(user=self.request.user)
+        user_role = user.role
+
+        if user_role == 'startup':
+            user_startups = UserStartup.objects.filter(customuser=self.request.user).values('startup')
+            return Startup.objects.filter(id__in=user_startups)
+        elif user_role == 'investor':
+            user_investors = UserInvestor.objects.filter(customuser=self.request.user).values('investor')
```
mehalyna: When building the queryset, using `values_list('startup', flat=True)` would be more appropriate and efficient than `values('startup')`. The same for `values('investor')`

ForumProject/users/views.py
```diff
@@ -60,6 +66,76 @@ def post(self, request, *args, **kwargs):
...
+                    return Response({'success': 'The startup company was successfully selected'}, status=status.HTTP_200_OK)
+            return Response({'error': 'You must select only your company'}, status=status.HTTP_400_BAD_REQUEST)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+class UserCompanyView(generics.ListAPIView):
+    permission_classes = []  # write permission only for users with role
+
+    def get_serializer_class(self):
+        user = UserRoleCompany.objects.get(user=self.request.user)
```
mehalyna: Similar to `CompanySelectionView`, ensure that exceptions are handled if `UserRoleCompany.objects.get(user=self.request.user)` fails

mehalyna (changes requested):

ForumProject/investors/serializers.py
Andrewakiv: Done, thanks. I forgot about this simple rule.

ForumProject/investors/views.py
Andrewakiv: My bad. Thanks. This simple class greatly reduces the amount of code.

ForumProject/investors/views.py
Andrewakiv: Done.

ForumProject/projects/views.py
Andrewakiv: Done.

ForumProject/projects/views.py
Andrewakiv: Done.

ForumProject/startups/views.py
Andrewakiv: Done.

ForumProject/startups/views.py
Andrewakiv: Done.

ForumProject/users/views.py
Andrewakiv: Thanks, very useful advice.

ForumProject/users/views.py
Andrewakiv: I leave this method for now, but later I will change it, because there is still no functionality for creating a relationship  UserRoleCompany during registration

ForumProject/users/views.py
Andrewakiv: Thanks,  very useful advice.

ForumProject/users/views.py
Andrewakiv: Done, thanks.

ForumProject/users/views.py
Andrewakiv: Thanks. Good idea.

ForumProject/users/views.py
Andrewakiv: Done

mehalyna (approved):

PR #83, opened 2024-05-03
71# update database schema for saved profiles[23-1]

### LINKED ISSUE #71: Update Database Schema for Saved Profiles [23-1]
**Description:**
Update the database schema to include a new model or table that links investors with the startups they save or follow. This will involve creating a many-to-many relationship since an investor can follow multiple startups and a startup can be followed by multiple investors.

**Technical Steps:**
1. Define a new model `InvestorStartup` with fields for `investor_id` and `startup_id`.
2. Use Django’s `models.ForeignKey` to link to the `Investor` and `Startup` models respectively.
3. Add a `DateTimeField` to track when the investor saved the startup.
4. Run migrations to update the database schema.

Task  #21

DariyBautista:


ForumProject/subscriptions/models.py
```diff
@@ -0,0 +1,27 @@
...
+    Represents the relationship between an investor and a startup they follow.
+
+    Attributes:
+        investor (Investor): The investor following the startup.
+        startup (Startup): The startup being followed by the investor.
+        saved_at (DateTimeField): The datetime when the investor saved the startup.
+    """
+
+    investor = models.ForeignKey(Investor, on_delete=models.CASCADE)
+    startup = models.ForeignKey(Startup, on_delete=models.CASCADE)
```
mehalyna: The use of `models.ForeignKey` with `on_delete=models.CASCADE` is appropriate here, as it helps maintain data integrity by preventing orphan records. However, consider if this is the desired behavior, as sometimes you might want to keep historical data for reporting or analytics. In such cases, you might opt for `models.SET_NULL` with the fields set to allow nulls, or another strategy based on business needs of the project

ForumProject/subscriptions/models.py
mehalyna, line 1: Your model does not explicitly prevent duplicate entries. Depending on the application logic, you might want to ensure that the same investor cannot follow the same startup more than once. Add a `unique_together` constraint in the model's `Meta` class to ensure that each combination of investor and startup is unique:
```
class Meta:
    unique_together = [['investor', 'startup']]
```

mehalyna (changes requested):

ForumProject/subscriptions/models.py
DariyBautista: Done, thanks!

ForumProject/subscriptions/models.py
DariyBautista: Done, thanks!

mehalyna (approved):

PR #84, opened 2024-05-05
#63 Implement Notification Preference Settings

oleksandr-liashenko:


ForumProject/investors/serializers.py
```diff
@@ -34,4 +34,27 @@ class Meta:
...
+
+        Raises:
+            serializers.ValidationError: If the investor name is empty or not unique.
+        """
+
+        value = value.strip()
+        value = value[0].upper() + value[1:]
+        if not value:
+            raise serializers.ValidationError("Investor name cannot be empty.")
+        if Investor.objects.filter(investor_name=value).exists():
```
mehalyna: Your code checks if `value` is non-empty after modifying it but does not correctly handle cases where the entire string could be whitespace. For example, if `value` consists only of whitespace (`' '`), it will be empty after the `strip()`, and trying to access `value[0].upper() + value[1:`] will raise an `IndexError`

ForumProject/investors/serializers.py
```diff
@@ -34,4 +34,27 @@ class Meta:
...
+            value (str): The investor name to validate.
+
+        Returns:
+            str: The validated investor name.
+
+        Raises:
+            serializers.ValidationError: If the investor name is empty or not unique.
+        """
+
+        value = value.strip()
```
mehalyna: Apply a more thorough sanitization to ensure names with redundant internal spaces are treated uniformly. You might want to replace multiple spaces with a single one:
```
value = re.sub(r'\s+', ' ', value).strip()
```

ForumProject/investors/serializers.py
mehalyna: The check for uniqueness does not account for case sensitivity. Two names that differ only in their case (e.g., "Alice" vs. "alice") will be treated as different by default in some databases. Ensure consistency in how names are stored and checked, potentially using case-insensitive queries:
```
if Investor.objects.filter(investor_name__iexact=value).exists():
    raise serializers.ValidationError("Investor name must be unique.")
```

ForumProject/notifications/models.py
```diff
@@ -0,0 +1,30 @@
...
+        ('project budget changed', 'project budget changed'),
+        ('message sent', 'message sent'),
+    ]
+    INITIATOR_CHOICES = [
+        ('investor', 'investor'),
+        ('project', 'project'),
+    ]    
+    project = models.ForeignKey(Project, on_delete=models.SET_NULL, related_name='notice_on_project')
+    startup_name = models.CharField(max_length=Startup._meta.get_field('startup_name').max_length)
+    investor = models.ForeignKey(Investor, on_delete=models.SET_NULL, related_name='notice_to_investor')
```
mehalyna: Is it okay for notifications to exist without a linked project or investor (e.g., if they are deleted)?

ForumProject/notifications/models.py
```diff
@@ -0,0 +1,30 @@
...
+    trigger = models.CharField(max_length=50, choices=TRIGGER_CHOICES)
+    initiator = models.CharField(max_length=8, choices=INITIATOR_CHOICES)
+    date_time = models.DateTimeField(auto_now_add=True)
+
+    def __str__(self):
+        addressee = f'Startup {self.startup_name}' if self.initiator == 'investor' else f'Investor {self.investor.investor_name}'
```
mehalyna: You assume that related objects (like investor and project) and their fields (`investor_name`, `project_name`) are not `None`, which might not be the case given that both project and investor can be `NULL`

ForumProject/notifications/views.py
```diff
@@ -0,0 +1,25 @@
...
+        in_app_notifications = request.data.get('in_app_notifications', False)
+
+        try:
+            startup = Startup.objects.get(user=user)  # Assuming Startup model has a ForeignKey to User model
+            startup.email_notifications = email_notifications
+            startup.in_app_notifications = in_app_notifications
+            startup.save()
+            return Response({'message': 'Notification preferences updated successfully.'}, status=status.HTTP_200_OK)
+        except Startup.DoesNotExist:
+            return Response({'error': 'Startup not found.'}, status=status.HTTP_404_NOT_FOUND)
```
mehalyna: Add a meaningful error response if the user does not have an associated startup. If necessary, consider checking user permissions to ensure they are allowed to update notification preferences

ForumProject/notifications/views.py
```diff
@@ -0,0 +1,25 @@
...
+from rest_framework import status
+from .models import Startup
+
+
+class NotificationPreferencesAPIView(APIView):
+    def post(self, request):
+        user = request.user  # Assuming user authentication is implemented
+        # Assuming request data is in JSON format with keys 'email_notifications' and 'in_app_notifications'
+        email_notifications = request.data.get('email_notifications', False)
+        in_app_notifications = request.data.get('in_app_notifications', False)
```
mehalyna: The code fetches the values of `email_notifications` and `in_app_notifications` using `.get()` with a default of `False`. Ensure that the incoming values are explicitly cast to `bool` if you're expecting strictly Boolean values

ForumProject/notifications/views.py
```diff
@@ -0,0 +1,25 @@
...
+
+        try:
+            startup = Startup.objects.get(user=user)  # Assuming Startup model has a ForeignKey to User model
+            startup.email_notifications = email_notifications
+            startup.in_app_notifications = in_app_notifications
+            startup.save()
```
mehalyna: The `try` block catches any exception and returns the error message. This could expose sensitive information and may not provide a good user experience.
Limit exception handling to more specific errors and add proper logging to diagnose unexpected issues

ForumProject/notifications/views.py
mehalyna: Instead of catching a `Startup.DoesNotExist` error separately and returning `HTTP_404_NOT_FOUND`, consider checking if the startup exists before trying to save:
```
startup = Startup.objects.filter(user=user).first()
if not startup:
    return Response({'error': 'Startup not found.'}, status=status.HTTP_404_NOT_FOUND)
```

ForumProject/projects/serializers.py
```diff
@@ -16,10 +14,29 @@ class ProjectSerializer(serializers.ModelSerializer):
...
+            value (str): The project name to validate.
+
+        Returns:
+            str: The validated project name.
+
+        Raises:
+            serializers.ValidationError: If a project with the same name already exists for the startup.
+        """
+        
+        if Project.objects.filter(startup=self.initial_data['startup'], project_name=value).exists():
```
mehalyna: It would be best to trim leading/trailing spaces to avoid false uniqueness errors. Consider normalizing the capitalization to ensure consistent comparison (e.g., converting everything to lowercase):
```
value = value.strip().lower()
```

mehalyna: Accessing `self.initial_data['startup']` directly might result in a `KeyError` if the key does not exist. Use `.get()` to safely access `startup`

ForumProject/startups/serializers.py
mehalyna, line 1: Consider previous recommendations: 
 - Trimming leading and trailing spaces is good practice
 - Capitalizing just the first letter might not cover all cases, especially if the first character is not alphabetical or if there are multiple words.
 - Checking if value is empty should come before modifying capitalization.

ForumProject/startups/views.py
```diff
@@ -194,5 +194,48 @@ def get_queryset(self):
...
+        try:
+            startup = Startup.objects.get(customuser=user)
+            startup.email_notifications = email_notifications
+            startup.in_app_notifications = in_app_notifications
+            startup.save()
+            return Response({'message': 'Notification preferences updated successfully.'}, status=status.HTTP_200_OK)
+        except Startup.DoesNotExist:
+            return Response({'error': 'Startup not found.'}, status=status.HTTP_404_NOT_FOUND)
+        except Exception as e:
+            return Response({'error': str(e)}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
mehalyna: Catching a generic `Exception` is not always informative and may hide specific issues.

mehalyna (changes requested):

ForumProject/investors/serializers.py
oleksandr-liashenko: It's a rudimentary part of the code, we don't need it anymore.

ForumProject/investors/serializers.py
oleksandr-liashenko: It's a rudimentary part of the code, we don't need it anymore.

oleksandr-liashenko: It's a rudimentary part of the code, we don't need it anymore.

ForumProject/startups/views.py
oleksandr-liashenko: updated

ForumProject/notifications/models.py
oleksandr-liashenko: It will be changed in another branch

ForumProject/startups/serializers.py
oleksandr-liashenko: done

ForumProject/projects/serializers.py
oleksandr-liashenko: done

oleksandr-liashenko: done

ForumProject/notifications/models.py
oleksandr-liashenko: done

ForumProject/notifications/views.py
oleksandr-liashenko: done

oleksandr-liashenko: done

ForumProject/notifications/views.py
oleksandr-liashenko: done

mehalyna (approved):

ForumProject/notifications/views.py
oleksandr-liashenko: done

mehalyna (approved):

DmAlexx (approved):

DariyBautista (approved):

avu-ua:
Closed as a new pull request with combined code covering previous pull requests #63, #66 and #88 has been opened

PR #86, opened 2024-05-07
Implemented signals to detect when an investor follows a project or s…

### LINKED ISSUE #62: Implement Notification Trigger Mechanism [21-2 Notification Trigger Mechanisms & Services]
**Description:**
Develop backend mechanisms that trigger notifications based on specific actions by an investor, such as following a startup's profile or sending a message.

**Acceptance Criteria:**
- Implement event listeners or hooks within the application that detect when an investor follows a profile or sends a message.
- Upon detecting such events, a notification record should be created in the database.
- Ensure that the notification mechanism is efficient and does not introduce significant latency to the investor's actions.

Task #21

avu-ua:
…ubscribes for a share in it.


ForumProject/notifications/signals.py
```diff
@@ -0,0 +1,70 @@
...
+    Returns:
+    - Notification: The created `Notification` instance.
+    '''
+    Notification.objects.create(
+            project=instance.project,
+            startup=instance.project.startup,
+            investor=instance.investor,
+            trigger='follower(s) list changed' if follow_status_change else 'subscription changed',
+            initiator='investor',
+        )
```
mehalyna: Consider adding validation to ensure all fields (`project`, `startup`, `investor`, etc.) are present and valid

ForumProject/notifications/signals.py
mehalyna, line 1: 1) If additional metadata (like `timestamp`) is required, pass relevant parameters directly to the notification creation function.
2) Consider implementing bulk notification creation if multiple changes are processed simultaneously.

mehalyna (changes requested):

ForumProject/notifications/signals.py
avu-ua: Done

ForumProject/notifications/signals.py
avu-ua: On point 1 - class Notification does have a `models.DateTimeField`, however it has parameter `auto_now_add=True`, therefore I do not think passing any parameters to the notification creation function is required.
Point 2 - done.

mehalyna (approved):

PR #89, opened 2024-05-09
Implemented emailing of notifications to Startup owners

### LINKED ISSUE #88: Implement emailing of Notification(s) once triggered [21-2(1)]

avu-ua:


ForumProject/notifications/signals.py
```diff
@@ -44,42 +45,68 @@ def create_notifications(instances, follow_status_change=True):
...
+      (default: True). If False, it is considered a change in the subscription share.
+    """
+    #trigger recording of notifications to the database
+    notifications = create_notifications(instance, follow_status_change=follow_status_change)
+    
+    for notification in notifications:
+        # Get all CustomUsers' emails associated with this startup
+        project = instance.project
+        startup = project.startup
+        user_startups = UserStartup.objects.filter(startup=startup)
```
mehalyna: This line is run inside a loop for each notification. This could lead to performance issues due to repeated database queries. If possible, this query should be moved outside the loop or optimized to fetch all necessary data in fewer queries

ForumProject/notifications/signals.py
```diff
@@ -44,42 +45,68 @@ def create_notifications(instances, follow_status_change=True):
...
+        recipients = [user.customuser.email for user in user_startups]
+
+        if recipients:
+            # prepare data for email
+            change = 'follower(s) list' if follow_status_change else 'subscription'
+            subject = f'The Project "{project.name}" has a change in {change}.'
+            message = f'{str(notification)}.\n\n\n{str(instance)}'
+            # send email
+            send_notification_email(subject, message, recipients)
```
mehalyna: 1) There is no error handling for potential issues such as database errors or failures in sending emails. Implementing try-except blocks to manage exceptions such as sending emails or database operations failing could improve the robustness of the function.
2) The function sends emails inside a loop, which can be inefficient and slow if there are many recipients or notifications. Consider batching these operations or using asynchronous tasks to handle email sending to avoid blocking the main execution thread and improve performance.
_Example:_
```
# Move database query outside of the loop
project = instance.project
startup = project.startup
user_startups = UserStartup.objects.select_related('customuser').filter(startup=startup)
recipients = [user.customuser.email for user in user_startups if user.customuser.email]
```
3) The function assumes that `user.customuser.email` is always valid and available. Validation should be added to ensure that emails are in a proper format before attempting to send them.
4) If the number of users or notifications grows, this function could become a bottleneck. Consider using more scalable approaches such as message queues (e.g., [Celery ](https://docs.celeryq.dev/en/stable/django/index.html) with RabbitMQ or Redis) for managing notifications and email dispatch 
_Example:_
```
from celery import shared_task

@shared_task
def send_email_task(subject, message, recipients):
    send_notification_email(subject, message, recipients)
```
5) The function has redundant code constructs like converting notification and instance objects to strings multiple times. You could optimize by storing the string representation once and reusing it

ForumProject/notifications/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    - message (str): The body content of the email.
+    - recipient_list (list): List of email addresses to send to.
+    """
+    try:
+        send_mail(
+            subject,
+            message,
+            settings.EMAIL_HOST_USER,
+            recipient_list,
+            fail_silently=False
```
mehalyna: This relation will halt execution. Depending on how critical email sending is to your application's flow, you might want to handle this differently.
With `fail_silently=False`, any error in sending mail will raise an exception. While this is handled by the `try-except` block, you could consider more nuanced error handling or recovery strategies, especially in cases where only some mails fail to send (e.g., one of the recipient's email server is down)

ForumProject/notifications/utils.py
```diff
@@ -0,0 +1,25 @@
...
+        )
```
mehalyna: Sending emails directly in a request-response cycle can significantly slow down response times. If recipient_list contains many emails, this will block the server process until all emails are sent

ForumProject/notifications/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    except Exception as e:
+        logger.error(f"Failed to send email to {len(recipient_list)} recipients: {str(e)}")
```
mehalyna: Extend the logging to include more details about the error or possibly the recipients in case of failure, while ensuring not to log sensitive information, e.g.
```
logger.error(f"Failed to send email: {str(e)}; Subject: {subject}; Recipients: {recipient_list}")
```

ForumProject/users/admin.py
```diff
@@ -2,8 +2,20 @@
 
 from .models import CustomUser, UserInvestor, UserStartup, UserRoleCompany
 
-# Register your models here.
-admin.site.register(CustomUser)
-admin.site.register(UserInvestor)
-admin.site.register(UserStartup)
-admin.site.register(UserRoleCompany)
+
+class CustomUserAdmin(admin.ModelAdmin):
```
mehalyna, line 6: Add `list_filter` to improve the ability to filter records based on relevant fields, such as status flags or related model attributes:
```
class CustomUserAdmin(admin.ModelAdmin):
    list_display = ['id', 'email', 'first_name', 'last_name', 'is_active']
    list_editable = ['is_active']
    list_filter = ['is_active', 'email']
```
You should Implement `search_fields` in your admin to allow quick searching, especially useful for user management.
```
class CustomUserAdmin(admin.ModelAdmin):
    search_fields = ['email', 'first_name', 'last_name']
```

ForumProject/users/admin.py
```diff
@@ -2,8 +2,20 @@
...
+    list_display = ['id', 'email', 'first_name', 'last_name', 'is_active']
+    list_editable = ['is_active']
+admin.site.register(CustomUser, CustomUserAdmin)
+
+class UserInvestorAdmin(admin.ModelAdmin):
+    list_display = ['customuser', 'investor', 'investor_role_id']
+admin.site.register(UserInvestor, UserInvestorAdmin)
```
mehalyna, line 13: Consider using `select_related` or `prefetch_related` in the `get_queryset` method to optimize database access for foreign key fields
```
class UserInvestorAdmin(admin.ModelAdmin):
    def get_queryset(self, request):
        return super().get_queryset(request).select_related('customuser', 'investor')
```

ForumProject/users/admin.py
mehalyna, line 1: 1) If the models have a large number of records, displaying related fields directly (such as `customuser`, `investor`, `startup` admin models) could lead to performance issues due to database join operations.
2) The admin interface lacks search and filter capabilities on potentially large datasets, which could hinder the ability to efficiently manage records
(Pay attention on further examples)
3) Consider using Django's `inlines` to manage related data within the same page. This is particularly useful for managing relationships like those between users and roles or startups

mehalyna (changes requested):

ForumProject/notifications/signals.py
avu-ua: Done

ForumProject/notifications/signals.py
avu-ua: 1. Database errors handling has been added.
2. Emailing is now made asynchronously using threading.
3. Here I rely on basic email validation of the CustomUser  `modelsEmailField` field, and on the fact that this field is mandatory for a CustomUser instance creation. Given the project scope does not provide for a more rigorous email validations (like checking if the domain is valid) and the active status of the user's email gets checked by way of sending a registration link during the registration procedure, I think repeated validation of email address here would be an unjustified belt-n-braces approach.
4. Asynchronous sending of email, per point 2 above, addresses this point. 
5. The string representations are store once and reused.

ForumProject/notifications/utils.py
avu-ua: Email sending is now moved to an asynchronous process to avoid causing slowdowns. Also, logging has been enhanced.

ForumProject/notifications/utils.py
avu-ua: Please see above

ForumProject/notifications/utils.py
avu-ua: Please see above

ForumProject/users/admin.py
avu-ua: I slightly improved representation of Users on admin panel by introducing some changes to **users**/admin.py for my personal convenience, for the purposes of managing and tracking the database records during testing / debugging exercise of the Notifications app functionality. Whilst there is indeed a room for perfection of **users**/admin.py, I would be grateful if this pull request could be approved with users/admin.py being "as is", which is out of scope of the current task.

ForumProject/users/admin.py
avu-ua: As above

ForumProject/users/admin.py
avu-ua: As above

mehalyna (approved):

DariyBautista (approved):

avu-ua:
Closed as a new pull request with combined code covering previous pull requests #63, #66 and #88 has been opened

PR #92, opened 2024-05-10
#87 fix of permissions for startup endpoints 

### LINKED ISSUE #87: fix of permissions for startup endpoints

DariyBautista:
added new permission IsStartupMember, fixed permissions for startup endpoints, fixed critical errors in displaying subscriptions from deleted startups

### DIFF
--- a/ForumProject/startups/views.py
+++ b/ForumProject/startups/views.py
@@ -2,11 +2,10 @@ from django.shortcuts import render, get_object_or_404
 from rest_framework import viewsets, status, generics
 from rest_framework.permissions import IsAuthenticated, IsAdminUser
 from rest_framework.response import Response
-from rest_framework.exceptions import PermissionDenied
+from rest_framework.exceptions import PermissionDenied, ValidationError, NotFound
 from projects.models import Project
 from users.permissions import IsStartupRole, IsInvestorRole, IsStartupCompanySelected, IsCompanyMember, IsStartupMember
 from rest_framework.permissions import IsAuthenticated
-from rest_framework.exceptions import NotFound
 from .models import Startup
 from users.models import UserStartup
 from .serializers import StartupSerializer
@@ -64,7 +63,7 @@ class StartupViewSet(viewsets.ModelViewSet):
         try:
             obj = queryset.get(pk=pk)
         except Startup.DoesNotExist:
-            raise NotFound("No Startup matches the given query.")
+            raise NotFound(f"No Startup found for primary key {pk}.")
         self.check_object_permissions(self.request, obj)  # Check object permissions.
         return obj
     
@@ -94,7 +93,10 @@ class StartupViewSet(viewsets.ModelViewSet):
         """
         Handle requests to update a startup for a user.
         """
-        instance = self.get_object()
+        try:
+            instance = self.get_object()
+        except Startup.DoesNotExist:
+            return Response({"error": "Startup not found."}, status=status.HTTP_404_NOT_FOUND)
         
         # Check if the user is the owner of the startup
         user_startup = UserStartup.objects.filter(customuser=request.user, startup=instance)
@@ -102,8 +104,16 @@ class StartupViewSet(viewsets.ModelViewSet):
             return Response({"error": "You are not the owner of this startup."}, status=status.HTTP_403_FORBIDDEN)
         
         serializer = self.get_serializer(instance, data=request.data, partial=kwargs.pop('partial', False))
-        serializer.is_valid(raise_exception=True)
-        self.perform_update(serializer)
+        try:
+            serializer.is_valid(raise_exception=True)
+        except ValidationError as e:
+            return Response({"error": e.detail}, status=status.HTTP_400_BAD_REQUEST)
+        
+        try:
+            self.perform_update(serializer)
+        except Exception as e:
+            return Response({"error": f"An error occurred while updating the startup: {str(e)}"}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+        
         return Response(serializer.data)
 
 
@@ -119,9 +129,8 @@ class StartupViewSet(viewsets.ModelViewSet):
         if not user_startup:
             raise PermissionDenied("You are not the owner of this startup.")
         
-        # Checking whether the startup has open projects
-        projects = Project.objects.filter(startup_id=instance.id)
-        if any(project.status != 'closed' for project in projects):
+        # Checking whether the startup has open projects using a database query
+        if Project.objects.filter(startup_id=instance.id).exclude(status='closed').exists():
             raise PermissionDenied("Cannot delete startup with ongoing projects.")
         
         # If the startup has all projects closed and the user is the owner,
--- a/ForumProject/startups/views.py
+++ b/ForumProject/startups/views.py
@@ -4,9 +4,9 @@ from rest_framework.permissions import IsAuthenticated, IsAdminUser
 from rest_framework.response import Response
 from rest_framework.exceptions import PermissionDenied
 from projects.models import Project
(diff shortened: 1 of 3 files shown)


ForumProject/startups/views.py
```diff
@@ -38,16 +38,36 @@ def get_permissions(self):
...
 
+        Raises:
+            NotFound: If no startup matches the given query.
+        """
+        queryset = self.filter_queryset(self.get_queryset())
+        pk = self.kwargs.get('pk')
+        try:
+            obj = queryset.get(pk=pk)
+        except Startup.DoesNotExist:
+            raise NotFound("No Startup matches the given query.")
```
mehalyna: The error message is clear, but it might be enhanced by including the attempted primary key in the message to aid in debugging or logging, like "No Startup found for primary key {pk}."

ForumProject/startups/views.py
```diff
@@ -70,30 +90,42 @@ def create(self, request, *args, **kwargs):
...
+        
+        # Check if the user is the owner of the startup
+        user_startup = UserStartup.objects.filter(customuser=request.user, startup=instance)
+        if not user_startup.exists():
+            return Response({"error": "You are not the owner of this startup."}, status=status.HTTP_403_FORBIDDEN)
+        
+        serializer = self.get_serializer(instance, data=request.data, partial=kwargs.pop('partial', False))
+        serializer.is_valid(raise_exception=True)
+        self.perform_update(serializer)
+        return Response(serializer.data)
```
mehalyna: While the method handles the case where the user is not the owner, consider what happens if `get_object()` does not find any object (it should raise a 404). Also, handle potential exceptions that might arise during the save process in `perform_update(serializer)`

ForumProject/startups/views.py
```diff
@@ -70,30 +90,42 @@ def create(self, request, *args, **kwargs):
...
         # Checking whether the startup has open projects
+        projects = Project.objects.filter(startup_id=instance.id)
         if any(project.status != 'closed' for project in projects):
             raise PermissionDenied("Cannot delete startup with ongoing projects.")
         
-        # If the startup has all projects closed, then deletion is possible
+        # If the startup has all projects closed and the user is the owner,
+        # then deletion is possible
         instance.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
```
mehalyna: Consider the efficiency of loading all projects and checking their statuses. If the number of projects is large, this could be inefficient. You might consider using a database aggregate function to check the status directly in the database query, e.g.
```
if Project.objects.filter(startup_id=instance.id).exclude(status='closed').exists():
    raise PermissionDenied("Cannot delete startup with ongoing projects.")
```

ForumProject/users/permissions.py
```diff
@@ -34,6 +36,25 @@ class IsStartupRole(IsRoleSelected):
...
+            Returns:
+                bool: True if the user is logged in and has the role of an investor, False otherwise.
+            """
+            try:
+                if request.user.is_authenticated and hasattr(request.user, 'user_info') and request.user.user_info.role == 'startup':
+                    return True
+                else:
+                    return False
+            except AttributeError:
+                return Response({"error": "Please log in to access this resource."}, status=status.HTTP_403_FORBIDDEN)
```
mehalyna: The documentation and function name suggest that it checks for an investor role, but the actual implementation checks if the user has a 'startup' role. Ensure that the role check aligns with the intended use. If it's meant to check for an investor, the condition should compare against 'investor' instead of 'startup'.

mehalyna: The `try block` is used to catch `AttributeError`, which could occur if `request.user` or `request.user.user_info` doesn't have the expected attributes. However, returning a `Response` object directly from the permission class method is not standard practice and won't work as expected. Permission methods should return `True` or `False`.
Instead of catching `AttributeError`, a more Django-practice approach would be to safely access attributes with checks:
```
if request.user.is_authenticated and hasattr(request.user, 'user_info'):
    return request.user.user_info.role == 'startup'
return False
```

mehalyna: Permissions should not attempt to return `Response` objects. If you need to handle errors or exceptions, consider raising an appropriate exception like `PermissionDenied` from `rest_framework.exceptions`. For example:
```
if not request.user.is_authenticated:
    raise PermissionDenied({"error": "Please log in to access this resource."})
```

ForumProject/users/permissions.py
```diff
@@ -34,6 +36,25 @@ class IsStartupRole(IsRoleSelected):
...
+            """
+            Check if the user is logged in and has the role of an investor.
+
+            Args:
+                request: The request object.
+                view: The view object.
+
+            Returns:
+                bool: True if the user is logged in and has the role of an investor, False otherwise.
+            """
```
mehalyna: Update the docstring to accurately reflect what the function is checking and correct any discrepancies in the explanation vs. the actual code logic.

ForumProject/users/permissions.py
```diff
@@ -42,6 +63,25 @@ class IsInvestorRole(IsRoleSelected):
...
+            try:
+                if request.user.is_authenticated and hasattr(request.user, 'user_info') and request.user.user_info.role == 'investor':
+                    return True
+                else:
+                    return False
+            except AttributeError:
+                return Response({"error": "Please log in to access this resource."}, status=status.HTTP_403_FORBIDDEN)
```
mehalyna: Returning a `Response` object directly from a permission class method is not standard and will not work as expected. Instead, permission methods should return `True` or `False`.
If there's an attribute error, returning a `Response` object is not appropriate in the context of permission checks. You should either handle the lack of attributes gracefully or raise an appropriate exception if the error state is unexpected.

mehalyna: Maybe something like this instead of class `IsInvestorRole`?
```
class IsInvestorPermission(BasePermission):
    ROLE = 'investor'

    def has_permission(self, request, view):
        """
        Check if the user is logged in and has the role of an investor.
        Args:
            request: The request object.
            view: The view object.
        Returns:
            bool: True if the user is logged in and has the role of an investor, False otherwise.
        """
        if not request.user.is_authenticated:
            raise PermissionDenied({"error": "Authentication credentials were not provided."})

        # Use getattr to avoid AttributeError and provide a default if user_info does not exist
        user_role = getattr(request.user, 'user_info', None)

        if user_role and user_role.role == self.ROLE:
            return True
        return False
```

mehalyna (changes requested):

ForumProject/users/permissions.py
DariyBautista: Thanks for the variation, but I will leave it as it is, correcting the previous comments

DariyBautista: Done, thanks!

ForumProject/users/permissions.py
DariyBautista: Done!

ForumProject/users/permissions.py
DariyBautista: Done, thanks!

DariyBautista: Done, thanks!

DariyBautista: I apologize for my inattention
Done!

ForumProject/startups/views.py
DariyBautista: Done, thanks!

ForumProject/startups/views.py
DariyBautista: Done, thanks!

ForumProject/startups/views.py
DariyBautista: Done, thanks!

mehalyna (approved):

PR #94, opened 2024-05-12
#75-testing-and-integration-with-existing-features-[23-5]

### LINKED ISSUE #75: Testing and Integration with Existing Features [23-5]
**Description:**
Conduct comprehensive tests to ensure the new features work as expected and integrate seamlessly with existing features like the startup profile viewing and the broader notification system.

**Technical Steps:**
1. Write unit tests for each new model and endpoint.
2. Conduct integration tests to see how new features interact with existing ones.
3. Test for performance issues and ensure the system scales well with an increasing number of saved profiles and notifications.
4. Use tools like Postman or automated scripts to simulate real user interactions and check for edge cases and security vulnerabilities.

Task #23

DariyBautista:


ForumProject/subscriptions/tests.py
```diff
@@ -1,3 +1,386 @@
...
+
+        # Create startup profile
+        self.startup_profile = Startup.objects.create(
+            startup_name='Test Startup',
+            startup_industry='Tech',
+            startup_phone='+3809876543',
+            startup_country='UA',
+            startup_city='Test City',
+            startup_address='Test Address'
+        )
```
mehalyna, line 124: You are setting credentials multiple times for different users in `setUp()`, which might lead to unintended behavior or confusion about which user is authenticated in subsequent test methods. Make sure to set or reset credentials at the beginning of each test method if tests require specific user authentication

ForumProject/subscriptions/tests.py
mehalyna, line 1: - Avoid redundant operations like re-creating the same setup in every test unless it is necessary. This can slow down test execution. If multiple tests need the same objects, consider creating them once either in `setUpTestData` or using `setUp` in a way that checks if they already exist before creating new ones.
- Your tests should also cover negative scenarios and errors. For instance, try to subscribe a user with an invalid token, or update a startup profile without the required permissions, and assert that the system correctly handles these cases.
- Check not only HTTP status codes but also the contents of the responses where applicable. For example, when checking notifications, ensure the contents match expected results, not just the count.
- After creating test data, especially using POST methods in tests, ensure you have mechanisms to clean up or use Django's testing framework's rollback functionalities to prevent pollution of the test database.

mehalyna (changes requested):

ForumProject/subscriptions/tests.py
DariyBautista: added tearDown method,
unintended behavior or confusion is unlikely because the creation of each profile is accompanied by different tokens

Done, thanks!

ForumProject/subscriptions/tests.py
DariyBautista: Done, thanks!

mehalyna (approved):

PR #95, opened 2024-05-12
#76 setup django channels 19 1]

Andrewakiv:


ForumProject/communications/consumers.py
```diff
@@ -0,0 +1,110 @@
...
+        self.room_group_name = None
+        self.room = None
+        self.messages = None
+        self.user = None
+        self.user_inbox = None
+
+    def connect(self):
+        self.room_name = self.scope['url_route']['kwargs']['room_name']
+        self.room_group_name = f'chat_{self.room_name}'
+        self.room = Room.objects.get(name=self.room_name)
```
mehalyna: The line might raise a `Room.DoesNotExist` exception if the room name does not exist in the database. Consider handling this exception to either create the room or send an error message back to the client.

ForumProject/communications/consumers.py
```diff
@@ -0,0 +1,110 @@
...
+            self.room_group_name,
+            self.channel_name,
+        )
+
+        self.send(json.dumps({
+            'type': 'user_list',
+            'users': [user.first_name for user in self.room.online.all()],
+        }))
+
+        if self.user.is_authenticated:
```
mehalyna: You check `if self.user.is_authenticated` only after sending some initial data to the client. It's better to check user authentication at the very beginning of the `connect()` method to ensure only authenticated users can connect.

ForumProject/communications/consumers.py
```diff
@@ -0,0 +1,110 @@
...
+        self.room = None
+        self.messages = None
+        self.user = None
+        self.user_inbox = None
+
+    def connect(self):
+        self.room_name = self.scope['url_route']['kwargs']['room_name']
+        self.room_group_name = f'chat_{self.room_name}'
+        self.room = Room.objects.get(name=self.room_name)
+        self.messages = Message.objects.filter(room=self.room)
```
mehalyna: You are fetching messages, but these messages are only used if the user is authenticated. Consider moving this inside the authentication check block to avoid unnecessary database queries when the user is not authenticated.

ForumProject/communications/consumers.py
```diff
@@ -0,0 +1,110 @@
...
+                self.user_inbox,
+                self.channel_name,
+            )
+            async_to_sync(self.channel_layer.group_send)(
+                self.room_group_name,
+                {
+                    'type': 'users_messages',
+                    'messages': [str(message) for message in self.messages],
+                }
+            )
```
mehalyna: You send group messages (`users_messages`) right after adding the user to the group. Make sure the order of operations is correct: the user should be added to the group before broadcasting messages to ensure they receive all relevant communications.

ForumProject/communications/consumers.py
```diff
@@ -0,0 +1,110 @@
...
+                self.room_group_name,
+                {
+                    'type': 'user_leave',
+                    'user': self.user.first_name,
+                }
+            )
+            self.room.online.remove(self.user)
+
+    def receive(self, text_data=None, bytes_data=None):
+        text_data_json = json.loads(text_data)
```
mehalyna: Consider adding error handling for JSON parsing to catch and handle malformed data gracefully.

ForumProject/communications/consumers.py
mehalyna, line 1: You use `async_to_sync()` quite frequently. Be aware that this could introduce performance bottlenecks. If possible, consider implementation of your consumer using `AsyncWebsocketConsumer` to be fully asynchronous
https://channels.readthedocs.io/en/latest/tutorial/part_3.html

ForumProject/communications/static/room.js
```diff
@@ -0,0 +1,105 @@
...
+    let oldOption = document.querySelector("option[value='" + value + "']");
+    if (oldOption !== null) oldOption.remove();
+}
+
+// focus 'chatMessageInput' when user opens the page
+chatMessageInput.focus();
+
+// submit if the user presses the enter key
+chatMessageInput.onkeyup = function(e) {
+    if (e.keyCode === 13) {  // enter key
```
mehalyna: Using a named constant for keycodes can make the code more readable.

ForumProject/communications/static/room.js
```diff
@@ -0,0 +1,105 @@
...
+        chatLog.scrollTop = chatLog.scrollHeight;
+    };
+
+    chatSocket.onerror = function(err) {
+        console.log("WebSocket encountered an error: " + err.message);
+        console.log("Closing the socket.");
+        chatSocket.close();
+    }
+}
+connect();
```
mehalyna: The function and event handlers are too long. Breaking these down into smaller functions (e.g., one to handle incoming messages, another to manage the user list, etc.) can improve readability and maintainability.

ForumProject/communications/views.py
```diff
@@ -1,3 +1,26 @@
...
+def index_view(request):
+    users = User.objects.exclude(id=request.user.id)
+    rooms_list = Room.objects.filter(name__contains=str(request.user.id))
+    return render(request, 'index.html', {
+        'rooms': rooms_list, 'users': users
+    })
+
+
+def room_view(request, user_id):
+    user = get_object_or_404(User, id=user_id)
```
mehalyna, line 21: The `get_object_or_404` usage is appropriate for fetching the user based on the `user_id`. However, consider situations where the `user_id` is invalid or malformed to avoid server errors

mehalyna (changes requested):

ForumProject/communications/consumers.py
Andrewakiv: Done, I added try-except and logging for this case

ForumProject/communications/consumers.py
Andrewakiv: Done, thanks

ForumProject/communications/consumers.py
Andrewakiv: Done, thanks

ForumProject/communications/consumers.py
Andrewakiv: Thanks, I removed this code and resolved this issue by passing users messages from the view function to the template

ForumProject/communications/consumers.py
Andrewakiv: Done, I added try-except and logging for this case

ForumProject/communications/consumers.py
Andrewakiv: Done, thanks

ForumProject/communications/static/room.js
Andrewakiv: Done, thanks

ForumProject/communications/static/room.js
Andrewakiv: Done, thanks

ForumProject/communications/views.py
Andrewakiv: if the user tries to write to a person who does not exist, he receives a 404 error. 
Should it be the other way around?

mehalyna: I mean the case, when`user_id` isn't a valid identifier. Trying to fetch a user with such an ID using `get_object_or_404` will raise a `ValueError` or `TypeError` before it makes any database query. You may leave this line without changes. Or you may add something like this
```
    try:
        user_id = int(user_id)  
    except ValueError:
        return HttpResponse('Invalid user ID format.', status=400)
```
before it

mehalyna (approved):

Issue #78, opened 2024-05-03
Develop Messaging API Endpoints [19-3]


mehalyna:
**Description:**
Develop REST API endpoints for sending and receiving messages, leveraging Django REST Framework along with Django Channels for real-time capabilities.

**Technical Steps:**
1. Create endpoints for creating a new conversation.
2. Develop endpoints to send a message within a conversation.
3. Implement an endpoint to retrieve message history in a conversation.

E.g. **API Endpoints**:
   - *Create Conversation*:
     - Endpoint: `POST /api/conversations/`
     - Body: `{participants: [user_ids]}`
   - *Send Message*:
     - Endpoint: `POST /api/messages/`
     - Body: `{conversation_id: id, text: "message text"}`
   - *List Messages in a Conversation*:
     - Endpoint: `GET /api/conversations/{conversation_id}/messages`

Task #19

PR #96, opened 2024-05-12
#78 develop messaging api endpoints 19 3]

DmAlexx:


ForumProject/communications/consumers.py
```diff
@@ -0,0 +1,110 @@
...
+        self.room_group_name = None
+        self.room = None
+        self.messages = None
+        self.user = None
+        self.user_inbox = None
+
+    def connect(self):
+        self.room_name = self.scope['url_route']['kwargs']['room_name']
+        self.room_group_name = f'chat_{self.room_name}'
+        self.room = Room.objects.get(name=self.room_name)
```
mehalyna: The `Room.objects.get(name=self.room_name)` call  could raise a `Room.DoesNotExist` exception if the room does not exist. It's better to handle this case explicitly to prevent the WebSocket connection from being accepted if the room does not exist, which could lead to errors down the line.

ForumProject/communications/consumers.py
mehalyna, line 1: 1) You are checking if `self.user.is_authenticated` multiple times. It would be cleaner and more efficient to handle this once at the beginning of the `connect` method and `disconnect` or prevent further execution if the user is not authenticated.
2) When broadcasting messages, there's currently no filtering to ensure that a user can only send messages if they are a participant in the room. This could potentially allow a user to send messages to a room they are not part of if they can guess or know the room's name.
3) You add the user to the room group and potentially other groups without ensuring that the connect is actually successful from a business logic perspective (e.g., the room exists, and the user is allowed in the room). Ensure all checks are passed before adding the user to groups.

ForumProject/communications/consumers.py
```diff
@@ -0,0 +1,110 @@
...
+
+        async_to_sync(self.channel_layer.group_send)(
+            self.room_group_name,
+            {
+                'type': 'chat_message',
+                'user': self.user.first_name,
+                'message': message,
+            }
+        )
+        Message.objects.create(user=self.user, room=self.room, content=message)
```
mehalyna: The `Message.objects.create`  does not handle any exceptions that could occur during database operations. Consider adding exception handling around this to manage issues like database write failures gracefully.

ForumProject/communications/models.py
```diff
@@ -1,3 +1,45 @@
...
+            'user_1': int(self.name.split('_')[1:][0]),
+            'user_2': int(self.name.split('_')[1:][1])
+        }
+
+    def get_users_names(self):
+        User = get_user_model()
+        return {
+            'user_1': User.objects.get(id=self.get_users_id()['user_1']).first_name,
+            'user_2': User.objects.get(id=self.get_users_id()['user_2']).first_name
+        }
```
mehalyna, line 32: These methods assume a specific format for the room's name field that includes two user IDs separated by underscores. This is prone to errors if the format is not strictly followed. Consider adding error handling to manage cases where the name does not contain the expected number of underscores or user IDs.

ForumProject/communications/models.py
```diff
@@ -1,3 +1,45 @@
...
+        return {
+            'user_1': int(self.name.split('_')[1:][0]),
+            'user_2': int(self.name.split('_')[1:][1])
+        }
+
+    def get_users_names(self):
+        User = get_user_model()
+        return {
+            'user_1': User.objects.get(id=self.get_users_id()['user_1']).first_name,
+            'user_2': User.objects.get(id=self.get_users_id()['user_2']).first_name
```
mehalyna, line 31: The `User.objects.get()` call can raise a `User.DoesNotExist` exception if no user with the provided ID exists. Handling this case gracefully would improve the robustness of the method.

ForumProject/communications/static/room.js
```diff
@@ -0,0 +1,105 @@
...
+    if (oldOption !== null) oldOption.remove();
+}
+
+// focus 'chatMessageInput' when user opens the page
+chatMessageInput.focus();
+
+// submit if the user presses the enter key
+chatMessageInput.onkeyup = function(e) {
+    if (e.keyCode === 13) {  // enter key
+        chatMessageSend.click();
```
mehalyna: The usage of `e.keyCode` is deprecated. Consider using `e.key` instead for detecting the enter key press:
```
chatMessageInput.onkeyup = function(e) {
    if (e.key === "Enter") {
        chatMessageSend.click();
    }
};
```

ForumProject/communications/views.py
mehalyna, line 1: 1) Error handling is minimal. Consider adding more robust error handling and user feedback mechanisms.
2) Adding explicit permission classes to views handling sensitive data like messages and conversations is strongly advised to protect user data.
3) Consider adding more detailed docstrings to each view to explain the purpose, expected parameters, and return types, which can be beneficial for maintenance and future development.

mehalyna (changes requested):

mehalyna (approved):

Andrewakiv (approved):

PR #97, opened 2024-05-12
#80 notification system for new messages 19 5]

DmAlexx:


ForumProject/communications/consumers.py
mehalyna, line 1: It might be beneficial to add more specific error handling around database operations or network calls, which could fail due to various reasons.

ForumProject/communications/consumers.py
```diff
@@ -0,0 +1,135 @@
...
+            return
+
+        self.room_name = self.scope['url_route']['kwargs']['room_name']
+        self.room_group_name = f'chat_{self.room_name}'
+        self.room = await get_room(self.room_name)
+        self.messages = await get_messages(self.room)
+        self.user = self.scope['user']
+        self.user_inbox = f'inbox_{self.user.first_name}'
+
+        await self.accept()
```
mehalyna: You're accepting the connection after performing some operations (e.g., getting the room). It’s usually better to accept the connection as soon as you determine the user is valid to ensure a responsive user experience.

ForumProject/communications/consumers.py
```diff
@@ -0,0 +1,135 @@
...
+from channels.generic.websocket import AsyncWebsocketConsumer
+
+from .utils import get_room, get_messages, get_online_users, get_user_first_name, \
+    add_user_to_online, remove_user_from_online, create_message
+
+
+class ChatConsumer(AsyncWebsocketConsumer):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(args, kwargs)
```
mehalyna: The call to the superclass constructor appears incorrect. It should be `super().__init__(*args, **kwargs)`. The current implementation might cause errors because `arg`s and `kwargs` are being passed as tuple and dict objects rather than unpacked arguments.

ForumProject/communications/models.py
```diff
@@ -1,3 +1,82 @@
...
+#             message=instance
+#         )
+#         channel_layer = get_channel_layer()
+#         async_to_sync(channel_layer.group_send)(
+#             f'notifications_{instance.room.online.id}',
+#             {
+#                 'type': 'send_notification',
+#                 'notification': f'New message in {instance.room.name}'
+#             }
+#         )
```
mehalyna: It's not allowed to push commented lines of code into the repo

ForumProject/communications/signals.py
mehalyna, line 1: Both signal handlers are using the same signal `post_save` for `Message`. If both are meant to run under the same conditions, they should be combined into a single function to improve clarity and efficiency.

ForumProject/communications/signals.py
```diff
@@ -0,0 +1,32 @@
...
+            message=instance
+        )
+
+@receiver(post_save, sender=Message)
+def send_chat_notification_via_channels(sender, instance, created, **kwargs):
+    if created:
+        chat_notification = ChatNotification.objects.create(
+            recipient=instance.room.online,
+            message=instance
+        )
```
mehalyna: Check this case, pls. The handler attempts to assign the recipient of a `ChatNotification` to a queryset `instance.room.online`, which might be incorrect since the recipient field expects a single `User` instance. This may raise an error. You need to iterate over the users in `instance.room.online.all()` and create/send notifications individually.

ForumProject/communications/signals.py
```diff
@@ -0,0 +1,32 @@
...
+        channel_layer = get_channel_layer()
+        async_to_sync(channel_layer.group_send)(
+            f'chat_notifications_{instance.room.online.id}',
```
mehalyna: I'm not sure, that the code `instance.room.online.id` is valid because `instance.room.online` is a queryset, not a single instance. This needs to be revised based on how you intend to manage channel groups. Typically, each user would have their own notification channel group or a room-based channel group.

ForumProject/communications/static/room.js
```diff
@@ -0,0 +1,115 @@
...
+
+// focus 'chatMessageInput' when user opens the page
+chatMessageInput.focus();
+
+const ENTER_KEY_CODE = 13;
+// submit if the user presses the enter key
+chatMessageInput.onkeyup = function(e) {
+    if (e.keyCode === ENTER_KEY_CODE) {  // enter key
+        chatMessageSend.click();
+    }
```
mehalyna: Using the `keyCode` property is deprecated. It’s better to use` e.key` which is more readable

ForumProject/communications/static/room.js
mehalyna, line 1: There's no handling of cases where the WebSocket may not be connected when the send button is clicked. Adding a check to ensure the socket is open before sending a message could prevent errors:
```
if (chatSocket.readyState === WebSocket.OPEN) {
    chatSocket.send(JSON.stringify({ "message": chatMessageInput.value }));
    chatMessageInput.value = "";
} else {
    console.error("WebSocket is not open.");
}
```

ForumProject/communications/views.py
mehalyna, line 1: You are loading all messages of a room directly in the view. This can be inefficient if the number of messages is large. Consider paginating the messages or loading them asynchronously on the client side as needed.

ForumProject/communications/views.py
```diff
@@ -1,3 +1,50 @@
...
+def room_view(request, user_id):
+    user = get_object_or_404(User, id=user_id)
+    if request.user.id > user.id:
+        thread_name = f'chat_{request.user.id}_{user.id}'
+    else:
+        thread_name = f'chat_{user.id}_{request.user.id}'
+    chat_room, created = Room.objects.get_or_create(name=thread_name)
+
+    messages = Message.objects.filter(room=chat_room)
+    users_messages = '\n'.join([str(message) for message in messages])+'\n'
```
mehalyna: Using `'\n'.join([str(message) for message in messages])` to pass messages to the template is unusual and might not be the best way to handle message formatting. It's better to handle such formatting in the template itself or use JSON.

mehalyna (changes requested):

ForumProject/communications/models.py
oleksandr-liashenko: done

ForumProject/communications/signals.py
oleksandr-liashenko: done

ForumProject/communications/views.py
oleksandr-liashenko: added

ForumProject/communications/views.py
oleksandr-liashenko: done

mehalyna (approved):

Andrewakiv (approved):

PR #98, opened 2024-05-13
#63-66-88-revisited [Notification Preference]

avu-ua:
This branch implements changes that cover four sub-tasks:

1. Implement emailing of Notification(s) once triggered [21-2(1)]
2. Implement Notification Preference Settings [21-4 Preference Settings]
3. API Endpoints for Managing In-app Notification Preferences [21-5 In-app Notification]
4. API Endpoints for Managing Email Notification Preferences [21-6]


ForumProject/notifications/admin.py
```diff
@@ -15,3 +15,15 @@ def has_change_permission(self, request, obj=None):
...
+@admin.register(NotificationPreferences)
+
+class NotificationPreferencesAdmin(admin.ModelAdmin):
+    list_display = ['id',
+        'startup',
+        'email_on_followers_change',
+        'email_on_share_subscription',
+        'in_app_on_followers_change',
+        'in_app_on_share_subscription'
+    ]
```
mehalyna: The indentation and alignment of fields could be more consistent to improve code readability:
```
class NotificationPreferencesAdmin(admin.ModelAdmin):
    list_display = [
        'id',
        'startup',
        'email_on_followers_change',
        'email_on_share_subscription',
        'in_app_on_followers_change',
        'in_app_on_share_subscription'
    ]
```

ForumProject/notifications/models.py
```diff
@@ -54,4 +54,15 @@ class Meta:
...
+        ]
+
+class NotificationPreferences(models.Model):
+    startup = models.OneToOneField(Startup, 
+        on_delete=models.CASCADE,
+        related_name='notice_preferences',
+        null=True
+    )
+    email_on_followers_change = models.BooleanField(default=True, verbose_name='Followers change email notices')
+    email_on_share_subscription = models.BooleanField(default=True, verbose_name='Sibscriptions change email notices')
```
mehalyna: There's a minor spelling error in "Sibscriptions" which should be "Subscriptions"

ForumProject/notifications/serializers.py
```diff
@@ -1,9 +1,18 @@
...
         read_only_fields = ['project', 'startup', 'investor', 'trigger', 'initiator', 'date_time']
+
+class NotificationPreferencesSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = NotificationPreferences
+        fields = ['email_on_followers_change',
+            'email_on_share_subscription',
+            'in_app_on_followers_change',
+            'in_app_on_share_subscription'
+            ]
```
mehalyna: Your serializer does not include the startup field. Depending on the application's requirements, you might want to expose the startup relationship to allow read or write operations through the API. This would be especially important if the client needs to know which startup the preferences are associated with or if they need to set preferences for a specific startup.

ForumProject/notifications/serializers.py
mehalyna, line 1: Consider adding a docstring to the serializer to explain its use that other developers working on the project should know about.

ForumProject/notifications/signals.py
```diff
@@ -39,47 +49,83 @@ def create_notifications(instances, follow_status_change=True):
...
-    # Bulk create all notifications
-    if notifications:
-        return Notification.objects.bulk_create(notifications)
-    return []
+    try:
+        # Bulk create all notifications
+        if notifications:
+            return Notification.objects.bulk_create(notifications)
+    except DatabaseError as db_err:
+        logger.error(f"Database error while creating notifications: {str(db_err)}")
```
mehalyna, line 57: The function only logs the error without any further action like re-raising the error or handling it in a way that the caller can react appropriately. It's important to ensure that errors do not silently fail.

ForumProject/notifications/signals.py
```diff
@@ -39,47 +49,83 @@ def create_notifications(instances, follow_status_change=True):
...
         )
 
-    # Bulk create all notifications
-    if notifications:
-        return Notification.objects.bulk_create(notifications)
-    return []
+    try:
+        # Bulk create all notifications
+        if notifications:
+            return Notification.objects.bulk_create(notifications)
```
mehalyna, line 55: The `bulk_create operation` is not wrapped in a database transaction. This can potentially lead to partial writes. Consider using Django's `transaction.atomic()` to ensure that all operations are completed successfully or rolled back in case of failure.

ForumProject/notifications/signals.py
```diff
@@ -39,47 +49,83 @@ def create_notifications(instances, follow_status_change=True):
...
-    Parameters:
-    - sender (type): The model class that sent the signal (usually `InvestorProject`).
-    - instance (InvestorProject): The instance that triggered the signal.
-    - kwargs (dict): Additional arguments for the signal handler.
+    Args:
+    - sender (type): The model class that sent the signal.
+    - instance (InvestorProject): The InvestorProject instance that triggered the signal.
     """
-    create_notifications(instance)
+    do_notifications(instance)
```
mehalyna: Since the instance is deleted, the information needed to send notifications (like related `UserStartup` or email preferences) might not be available or relevant anymore. Consider the implications of sending notifications for deleted objects and ensure that all necessary data is still available.

ForumProject/notifications/signals.py
```diff
@@ -39,47 +49,83 @@ def create_notifications(instances, follow_status_change=True):
...
+            return Notification.objects.bulk_create(notifications)
+    except DatabaseError as db_err:
+        logger.error(f"Database error while creating notifications: {str(db_err)}")
 
-# Signal to create a notification when an investor starts following a project or changes a share
-@receiver(post_save, sender=InvestorProject)
-def investor_project_followed_or_subscription_changed(sender, instance, created, **kwargs):
+    return notifications
+
+def do_notifications(instance, follow_status_change=True):
```
mehalyna, line 61: The complexity in the `do_notifications` method can be reduced by separating concerns more distinctly (e.g., separating email sending logic into a different method)

ForumProject/notifications/views.py
mehalyna, line 1: The error messages and HTTP status codes are handled somewhat inconsistently. For instance, when a startup ID is not found, a `404 Not Found` is sometimes used, which is correct, but the initial check returns a 404 with a non-standard message. It would be more appropriate to use `400 Bad Request` when the request lacks necessary information (e.g., no startup selected).

mehalyna (changes requested):

ForumProject/notifications/admin.py
avu-ua: Done, thank you

ForumProject/notifications/models.py
avu-ua: Thank you, corrected

ForumProject/notifications/serializers.py
avu-ua: Done. Now the client gets startup_id with the response.

ForumProject/notifications/serializers.py
avu-ua: Done

ForumProject/notifications/signals.py
avu-ua: I understand that in this case the "caller" is ultimately an Investor who causes notification(s) indirectly, by subscribing for a Project's share or by changing the list of Projects he follows. The ultimate caller (Investor) is not supposed to be even aware of notifications being automatically triggered by his actions, and informing the Investor of any failure of a notification creation serves no purpose. Therefore at this stage no HTTP response to any user is envisaged that would report the issue, instead a log is proposed to be created to be available for the application's administrators.
If I misunderstood the definition of "caller", I would be grateful for a direction of what namely should be implemented in addition to the error logging.

ForumProject/notifications/signals.py
avu-ua: Here I, too, would ask for a permission to disagree - it seems to me that atomization of the bulk_create is not justified. If, for example, out of 20 notifications to be created in one go 1 fails for some reason, it should be incorrect to force cancellation of the other 19 notifications which are otherwise ready to be recorded in the DB and sent to their recipients. In other words, my view is that on this occasion the application logic should be "if one notification fails this should not affect the others" as opposed to "either all notifications should succeed or none".

ForumProject/notifications/signals.py
avu-ua: I understand that in Django signal handlers, when a signal post_delete is received by the handler function, the instance that triggered the signal is still available within the handler function, even though it has been deleted from the database., i.e. the post_delete signal gets executed after the instance has been deleted from the database but before the transaction is committed. On that basis, there are no adverse implications of sending notifications caused by deletion of InvestorProject instance despite that instance's attributes are called within the post_delete signal handler.
However, thanks to you drawing attention to this aspect I have discovered that the notification that gets sent to the addressee upon the InvestorProject instance deletion was misleading as it conveyed outdated (pre-deletion status) information. The post_delete signal handler has been updated to correctly notify the addressee about the Project having lost a follower (which is exactly what happens when InvestorProject gets deleted, according to the application's business logic).

ForumProject/notifications/signals.py
avu-ua: I gave it a thought and must admit I do not know how to implement the proposed concerns separation without adverse effect on the code readability. Quite a big number of arguments would need to be passed to such new _helper_ method designed to separate some concerns `do_notifications` deals with from others.
I also note that the `do_notifications` method already has two _helper_ functions called inside of it - `create_notifications` and `utils.send_email_async`.

ForumProject/notifications/views.py
avu-ua: Done

ForumProject/notifications/signals.py
mehalyna: Let's leave this part without changes. However we'll think about this issue

mehalyna (approved):

PR #101, opened 2024-05-16
Tests for projects app

avu-ua:


alona-ruda (approved):

ForumProject/projects/tests.py
```diff
@@ -1,3 +1,443 @@
...
+        if method.upper() == 'GET':
+            return client.get(reverse(url_name, kwargs={'pk': pk}), data, format='json')
+        elif method.upper() == 'PUT':
+            return client.put(reverse(url_name, kwargs={'pk': pk}), data, format='json')
+        elif method.upper() == 'DELETE':
+            return client.delete(reverse(url_name, kwargs=pk) if pk else 
+                                 reverse(url_name), data, format='json')
+        else:
+            return client.post(reverse(url_name, kwargs=pk) if pk else 
+                               reverse(url_name), data, format='json')
```
mehalyna: The DELETE and POST method checks need to handle the `pk` correctly. Currently, they use `kwargs=pk` which might cause issues if `pk` is not `None`.

ForumProject/projects/tests.py
```diff
@@ -1,3 +1,443 @@
...
+            action (str): The action to perform ('follow', 'subscribe' or 'delist_project').
+
+        Returns:
+            Response: The response object returned by the API.
+        """
+        return self.visit_endpoint(
+            f'projects:{action}',
+            self.tokens[token_owner.email],
+            'POST',
+            pk = pk
```
mehalyna: Ensure `pk` passed is consistent and correctly handled. Also, consider renaming `pk` to `kwargs` for clarity.

ForumProject/projects/tests.py
```diff
@@ -1,3 +1,443 @@
...
+        Returns:
+            CustomUser: The created user object.
+        """
+        return CustomUser.objects.create_user(
+            email='user_alien@user.com',
+            first_name='User',
+            last_name='Alien',
+            phone_number='+3806669966',
+            password='-Qwerty-3',
+            is_active=True
```
mehalyna: You're using hardcoded values for user creation. These should be dynamically generated to avoid conflicts with existing data.

ForumProject/projects/tests.py
```diff
@@ -1,3 +1,443 @@
...
+            if i == 1:
+                refresh = RefreshToken.for_user(user_alien)
+                token = str(refresh.access_token)
+            if i == 2:
+                UserRoleCompany.objects.create(user=user_alien, role='startup')
+            response = self.visit_endpoint('projects:project-detail', token, 'GET',pk=1)
+            if i == 1:
+                self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+            else:
+                self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
```
mehalyna: The loop iteration `for i in range(2)` logic is unclear. It checks `if i == 1` and `if i == 2`, but `range(2)` will only iterate over 0 and 1.

mehalyna (changes requested):

ForumProject/projects/tests.py
avu-ua: No issues occur when `pk` is not `None`, I have checked this empirically. Please note `pk` has been renamed to `kwargs` according to your next comment.

ForumProject/projects/tests.py
avu-ua: Done

ForumProject/projects/tests.py
avu-ua: I am not sure how can this be dinamically generated. Also, values for this and other users creation (lines 101-108 and 127-137) are certainly not causing any conflicts, and no other user is created in the course of the tests run.

ForumProject/projects/tests.py
avu-ua: Line `for i in range(2):` was my oversight, this has now been corrected to `for i in range(3):`, thank you.

ForumProject/projects/tests.py
mehalyna: While the current hardcoded values may not be causing conflicts now, dynamically generating test data can provide several benefits:
- As the test suite grows or other developers contribute more tests, dynamic data generation helps ensure that new tests won’t accidentally conflict with existing ones.
- Dynamic generation helps simulate a wider range of scenarios, making the tests more robust and comprehensive.
- It follows best practices in testing to avoid hardcoded values, reducing maintenance efforts in the long run.
Here’s an example of how we can dynamically generate user data in our tests:
```
import random
import string

def random_string(length=10):
    return ''.join(random.choices(string.ascii_lowercase + string.digits, k=length))

def random_email():
    return f'{random_string()}@example.com'

user_data = {
    'email': random_email(),
    'first_name': random_string(),
    'last_name': random_string(),
    'phone_number': f'+380{random.randint(100000000, 999999999)}',
    'password': '-Qwerty-3',
    'is_active': True
}

user = CustomUser.objects.create_user(**user_data)
```
This approach ensures that each user created during the tests is unique and avoids any potential conflicts with existing data, now or in the future.
Regarding lines 101-108 and 127-137, while these currently do not cause conflicts, applying dynamic data generation consistently across all tests can help maintain uniformity and prevent potential issues as the project evolves.

avu-ua: Thank you for the help, That's been implemented.

mehalyna (approved):

PR #106, opened 2024-05-19
Project app refactored

avu-ua:
1. Added functionality of logs creation when a file is added for the Project or deleted, or when file description is changed.
2. Amended mechanism for triggering logs - now is arranged vie event listeners at signals.py rather hen in custom `create` / `update` / `destroy` method in viewsets. 
3. Amended logic for creation folders on the server to store files of the Project. Previously folders were named using startup-slugs and Project-slugs - so long as names of Startups and/or Projects can change that logic could lead to vulnerability of getting files lost on he server. Now the folders are named based on Primary Key of the Startups and Projects.


ForumProject/projects/signals.py
mehalyna, line 1: Add error handling for cases where the related objects (e.g., `UserRoleCompany`) might not exist. This prevents your signal handlers from crashing if the data is inconsistent.

mehalyna (changes requested):

ForumProject/projects/signals.py
avu-ua: This has been done.

alona-ruda (approved):

mehalyna (approved):

oleksandr-liashenko (approved):

PR #118, opened 2024-05-23
#112 implement funding check for investor subscriptions

alona-ruda:


DariyBautista (approved):

ForumProject/projects/models.py
```diff
@@ -151,8 +157,23 @@ def clean(self):
...
+
+        Args:
+            project_id (int): The ID of the project.
+
+        Returns:
+            int: The total amount of funding.
+        """
+        total_funding = cls.objects.filter(
+            project_id=project_id).aggregate(total=models.Sum('share'))['total']
+        return total_funding or 0
```
mehalyna: Add type hints for method parameters and return values.
Add basic error handling to manage potential issues gracefully.
_Example_
```
    def get_total_funding(cls, project_id: int) -> float:
        """
        Calculate the total funding from all investors for a specific project.
        
        Args:
            project_id (int): The ID of the project.
        
        Returns:
            float: The total amount of funding.
        """
        total_funding = cls.objects.filter(
            project_id=project_id
        ).aggregate(total=models.Sum('share'))['total']
        
        return float(total_funding or 0)
```

mehalyna (changes requested):

ForumProject/projects/models.py
alona-ruda: Added

mehalyna (approved):

